PR #2, opened 2018-06-29
fix: tidy up package.json, add pre-commit hook for tests

darahayes:
This PR makes some very small, developer friendly changes,

* reorders some bits in package.json
* changes some of the npm commands
* puts a placeholder test command
* removes the nodemon.json config that tries to auto lint. This regularly causes crashes which is pretty bad from a developer experience perspective
* adds a precommit hook which runs tests - we can keep this pretty lightweight by running integration tests in a different command. (If we need)

Ping @aliok @david-martin

### DIFF
--- a/nodemon.json
+++ b/nodemon.json
@@ -1,0 +1,5 @@
+{
+  "events": {
+    "restart": "npm run format"
+  }
+}
--- a/nodemon.json
+++ b/nodemon.json
@@ -1,5 +1,0 @@
-{
-  "events": {
-    "restart": "npm run format"
-  }
-}
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   "homepage": "https://github.com/aerogear/aerogear-data-sync-server#readme",
   "main": "index.js",
   "scripts": {
-    "test": "npm run lint && echo 'need to add tests'",
+    "test": "echo 'need to add tests'",
     "start": "node ./server.js",
     "dev": "SCHEMA_FILE=./examples/schema.example.graphql DATA_SOURCES_FILE=./examples/data-sources.example.json RESOLVER_MAPPINGS_FILE=./examples/resolver-mappings.example.json QUERY_FILE=./examples/query.example.graphql nodemon ./server.js",
     "lint": "standard",
@@ -44,6 +44,7 @@
     "subscriptions-transport-ws": "^0.9.11"
   },
   "pre-commit": [
+    "lint",
     "test"
   ]
 }
--- a/package.json
+++ b/package.json
@@ -2,13 +2,6 @@
   "name": "data-sync-server",
   "version": "1.0.0",
   "description": "GraphQL based server for syncing data between clients",
-  "main": "index.js",
-  "scripts": {
-    "test": "standard --fix && mocha",
-    "start": "node ./server.js",
-    "start-dev": "SCHEMA_FILE=./examples/schema.example.graphql DATA_SOURCES_FILE=./examples/data-sources.example.json RESOLVER_MAPPINGS_FILE=./examples/resolver-mappings.example.json QUERY_FILE=./examples/query.example.graphql nodemon ./server.js",
-    "format": "standard --fix"
-  },
   "repository": {
     "type": "git",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -2,13 +2,6 @@
   "name": "data-sync-server",
```
david-martin: I've changed the repo name to this

package.json
```diff
@@ -2,13 +2,6 @@
...
-  "main": "index.js",
-  "scripts": {
-    "test": "standard --fix && mocha",
-    "start": "node ./server.js",
-    "start-dev": "SCHEMA_FILE=./examples/schema.example.graphql DATA_SOURCES_FILE=./examples/data-sources.example.json RESOLVER_MAPPINGS_FILE=./examples/resolver-mappings.example.json QUERY_FILE=./examples/query.example.graphql nodemon ./server.js",
-    "format": "standard --fix"
-  },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/aerogear/aerogear-data-sync-server.git"
```
david-martin: would you mind changed the repo url here

nodemon.json
```diff
@@ -1,5 +0,0 @@
-{
```
david-martin: any particular reason you're removing this?

My thinking when adding it was most people would be using the `dev` script with nodemon,
so as they change code, it's automatically formatted.

package.json
```diff
@@ -24,8 +17,17 @@
     "url": "https://github.com/aerogear/aerogear-data-sync-server/issues"
   },
   "homepage": "https://github.com/aerogear/aerogear-data-sync-server#readme",
+  "main": "index.js",
```
aliok: Can you change this while you're at it?

aliok:
Tested pre commit hook locally and it works.
Good to merge after addressing the minor comments.

package.json
darahayes: I see the reason but I was playing around some of the code and one of things that happens every once in a while is that standard encounters an issue that cannot be automatically fixed. One example which occurs often enough in local dev is unused variables.

When this happens the whole thing crashes a bunch of times before failing. I think it's much handier if we simply include linting as part of precommit hooks and as part of CI checks. Happy to keep it though if you prefer

nodemon.json
darahayes: I see the reason but I was playing around some of the code and one of things that happens every once in a while is that standard encounters an issue that cannot be automatically fixed. One example which occurs often enough in local dev is unused variables.

When this happens the whole thing crashes a bunch of times before failing. I think it's much handier if we simply include linting as part of precommit hooks and as part of CI checks. Happy to keep it though if you prefer

package.json
darahayes: sure!

nodemon.json
david-martin: Could we make the pre-commit hook run `standard --fix`?
If it's just linting (and warnings/errors), there's nothing stopping badly formatted code getting into a PR.
In my experience, badly formatted code getting into a PR can distract from the actual goal of a PR (bikeshedding and such).
Lets fix it automatically so the review and discussion stays focused on the code, not formatting.

darahayes: I would advise against it. This is kind of difficult and dangerous to do because you would need to run `git add` again before the commit happens. How do you know what to add?

Tools that magically add stuff to `git commit` in the background are really dangerous IMO. In my experience, developers often selectively commit things and this kind of approach would lead to accidental files being committed.

I don't think this would distract reviewers too much because there are two protections in place. 1. when the developer runs `git commit` they won't be able to commit unless lint issues are fixed. 2. if they skip the precommit hook, the CI job will still fail on the linting also. I think this does a pretty good job ensuring reviewers don't need to waste time on this stuff.

david-martin: > I would advise against it. This is kind of difficult and dangerous to do because you would need to run git add again before the commit happens. How do you know what to add?

Understood.

So why not save the developer time and just fix formatting issues automatically before they even get to the `git add` phase?
The most likely thing they'll do after the git pre-commit fails is run the auto fix command anyways, so we can save them a step by doing it automatically.

darahayes:
@david-martin I've added back in nodemon.json 😄

PR #3, opened 2018-06-29
feat: initial CircleCI config

darahayes:
Adds a CircleCI config that will allow docker release stuff.

The actual docker commands should be implemented in a separate PR

@david-martin I have two suggestions for the docker build commands.

Either implement them directly in `package.json` or create a `scripts` directory with the appropriate bash scripts that perform the right commands, then wire up the commands in `package.json` to the scripts

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,6 +27,7 @@ version: 2
       steps:
         - checkout
         - setup_remote_docker
+        - run: npm run release:validate
         - run: npm run docker:build:release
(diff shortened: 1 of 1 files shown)


david-martin:
> @david-martin I have two suggestions for the docker build commands.

I think this depends on the complexity of the commands needed.
If its a one liner, inline in package.json.
If its more than 1 command & starting to get comlpex, a simple bash script that's referenced in package.json would suffice

PR #5, opened 2018-06-29
feat: docker build & release scripts

david-martin:


darahayes:
@david-martin looks great! Going to pull down this branch and just try it out a little bit.

scripts/docker_build_release.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+RELEASE_TAG ?= $(CIRCLE_TAG)
```
darahayes: This looks like Makefile syntax. It should be this:

```
RELEASE_TAG=$CIRCLE_TAG
```

Also anywhere else where variables are being used, the syntax `$(VARIABLE_NAME)` needs to be `$VARIABLE_NAME`. With the first one, bash tries to evaulate whatever is inside as a command :)

scripts/docker_push.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+APP_NAME = data-sync-server
```
darahayes: Bash sadly does not allow spaces in variable declarations :)

It should be like this

```
APP_NAME=data-sync-server
```

david-martin:
@darahayes pushed a cleanup of the scripts. Apologies for the initial sloppiness.

I've tried out the build & push for latest tag.
https://hub.docker.com/r/aerogear/data-sync-server/tags/

We can do a manual release here now of the first version if you like?
(would circleci pick up on that if we just tag on this branch?)

darahayes:
@david-martin No problem! You normally can test out the releases by pushing a tag directly to a branch although I'm not sure it would pick it up in this case because the branch is on your fork.

I would say let's merge and then we can do a test run from master. I'm fairly confident it will work and if not we can clean up in a second PR (branched from origin).

PS: just one minor question. Do we want the `latest` tag to reflect what is currently on master? This means that latest is always considered unstable. I'm happy to defer to you, although in other places such as aerogear-app-metrics and ups-config-operator. We have used the `master` tag for all master builds and then for releases we use both `TAG` and `latest` if that makes sense.

david-martin:
> PS: just one minor question. Do we want the latest tag to reflect what is currently on master? This means that latest is always considered unstable. I'm happy to defer to you, although in other places such as aerogear-app-metrics and ups-config-operator, we have used the master tag for all master builds and then for releases we use both TAG and latest if that makes sense.

@darahayes Lets follow the convention already used.
I've updated the scripts to build/push master instead of latest.
Also, for pushing the release it will tag the the release image as `latest` before pushing both images.
I'll leave you to merge if it looks correct.

darahayes:
@david-martin feel free to mearge. You will be able to watch the master build (including pushing to dockerhub here) https://circleci.com/gh/aerogear/data-sync-server

PR #6, opened 2018-07-02
Initial code division

aliok:
JIRA: https://issues.jboss.org/browse/AEROGEAR-3435
Divide code into modules

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 .vscode/
+.idea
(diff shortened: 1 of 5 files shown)


david-martin:
@aliok I'm OK with the initial division of code you have.
Something to consider is how easy we make it to add new datasource/resolver combinations to the server, for ourselves and potentially for anyone using the server.
There will always be a set of resolvers that go with a data source, so maybe there's a neater way of wrapping those 2 up? (not for this PR though I think)

darahayes:
Let's merge this and then do the tests in a separate PR :)

aliok:
@david-martin I think we can refactor stuff continuously when we're developing the new features.
This PR is basically to reduce conflicts and make things unit testable.

PR #8, opened 2018-07-02
AEROGEAR-3435: Add initial unit tests and handle some unhandled base cases

aliok:
JIRA: https://issues.jboss.org/browse/AEROGEAR-3435

- Introduced Ava testing framework
- Added unit tests to where it makes sense.
- While writing tests for base cases, I noticed there are some base cases not handled. I added handling for them.
- Validate Handlebars templates in advance to fail early (was waiting until a query/mutation is executed to fail)
- No integration tests!

Verification:
1. Do run npm test
2. Do start the app, execute some stuff over Graphiql

### DIFF
--- a/server/lib/resolverMaker.js
+++ b/server/lib/resolverMaker.js
@@ -18,13 +18,34 @@ module.exports = function (dataSourceTypes, dataSourceClients, resolverMappingsJ
     }
 
     _.forEach(resolverMapping, (value, key) => {
+      if (_.isEmpty(value.dataSource)) {
+        throw new Error('Missing data source for mapping: ' + key)
+      }
+
+      if (_.isEmpty(value.requestMapping)) {
+        throw new Error('Missing request mapping for mapping: ' + key)
+      }
+
+      if (_.isEmpty(value.responseMapping)) {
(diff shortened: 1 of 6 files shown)


darahayes:
This is so awesome @aliok

server/lib/resolverMaker.js
```diff
@@ -33,9 +54,29 @@ module.exports = function (dataSourceTypes, dataSourceClients, resolverMappingsJ
 }
 
 function buildPostgresResolver (dataSourceClient, requestMapping, responseMapping) {
+  // use Handlebars.precompile to fail early during initialization
+  try {
+    Handlebars.precompile(requestMapping)
+  } catch (ex) {
+    console.error('Compilation error in requestMapping: ' + requestMapping)
+    console.error(ex)
+    throw new Error('Compilation error in requestMapping: ' + requestMapping)
```
david-martin: @aliok Great to have this.
I'd say the exception thrown by precompile should make its way back to the developer (for the Admin UI).
I'm not sure how best to do that yet, but something to think about.

aliok: Good point. Let's keep that in mind.

PR #10, opened 2018-07-02
refactor: simplify dataSourceParser and resolverMaker

darahayes:
This PR doesn't introduce any new features or functionality. Just does some refactor before I move onto splitting resolvers and data sources into their own file.

Data Source Definitions used to look like this:

```json

{
  "Query": {
    "readNote": {
      "dataSource": "p1",
      "requestMapping": "SELECT id, title, content, timestamp FROM \"Notes\" WHERE id='{{context.arguments.id}}';",
      "responseMapping": "{{toJSON context.result.[0]}}"
    }
  },
  "Mutation": {
    "createNote": {
      "dataSource": "p1",
      "requestMapping": "INSERT INTO \"Notes\" (\"id\",\"title\",\"content\",\"timestamp\") VALUES ('{{context.arguments.id}}','{{context.arguments.title}}','{{context.arguments.content}}',{{context.arguments.timestamp}}) RETURNING *;",
      "responseMapping": "{{toJSON context.result.[0]}}"
    }
  }
}
```

now they look like this:

```
{
  "readNote": {
    "type": "Query",
    "dataSource": "p1",
    "requestMapping": "SELECT id, title, content, timestamp FROM \"Notes\" WHERE id='{{context.arguments.id}}';",
    "responseMapping": "{{toJSON context.result.[0]}}"
  },
  "createNote": {
    "type": "Mutation",
    "dataSource": "p1",
    "requestMapping": "INSERT INTO \"Notes\" (\"id\",\"title\",\"content\",\"timestamp\") VALUES ('{{context.arguments.id}}','{{context.arguments.title}}','{{context.arguments.content}}',{{context.arguments.timestamp}}) RETURNING *;",
    "responseMapping": "{{toJSON context.result.[0]}}"
  }
}
```

This allows for some simplification of the code in `resolverMaker.js`

I also changed the dataSourceParser to return a single object.

It used to look something like this:

```
{
	dataSourceTypes: {
		'p1': 'postgres'
	},
	dataSourceClients: {
		'p1': <Postgres Client Object>
	}
}
```

Now it looks like this:

```
{
	'p1': {type: 'postgres', client: <Client Object>}
}
```

The result of this is that the resolverMakerFunction signature has changed from this:

```
resolverMaker(dataSourceTypes, dataSourceClients, resolverMappings)
```

to this:

```
resolverMaker(dataSources, resolverMappings)
```

I was kind of struggling with mentally grasping the code and all of the objects and their relations before. I think this will really help going forward with the refactor to be done in this ticket: https://issues.jboss.org/browse/AEROGEAR-3436

I've also updated the appropriate tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,3 +85,30 @@ The baseline architecture is shown below:
 6. Configured Authentication & Autohorizatin checks are applied
 7. Logging & Metrics data is gathered from the Server & connected Clients
 
+## Running Tests
+
+```
+npm run test
+```
+
+### Debugging Individual Tests
+
+The easiest way to debug tests is using Chrome DevTools. Use [inspect-process](https://npm.im/inspect-process) to easily launch a debugging session with Chrome DevTools.
+
+```
+npm install -g inspect-process
+```
+
+* In chrome go to [`chrome://inspect`](chrome://inspect/)
+* Click on 'Open dedicated DevTools for Node.' This will open a new DevTools window.
+* Click on 'add folder to workspace' and use the wizard to open this project.
+* Go to the appropriate test file (or code that's being tested) and set a breakpoint
+* Now run the individual test as follows:
+
+```
+inspect node_modules/ava/profile.js some/test/file.js
+```
+
+The DevTools window should automatically connect to the debugging session and execution should pause if some breakpoints are set.
+
+
--- a/server/lib/resolverMaker.js
+++ b/server/lib/resolverMaker.js
@@ -32,20 +32,18 @@ module.exports = function (dataSources, resolverMappings) {
       throw new Error('Unknown data source "' + resolverMapping.dataSource + '" for mapping ' + resolverMappingName)
     }
 
-    if (['Query', 'Mutation'].indexOf(resolverMapping.type) < 0) {
-      return
-    }
-
     let { type, client } = dataSources[resolverMapping.dataSource]
 
     if (type === 'postgres') {
       try {
+        resolvers[resolverMapping.type] = resolvers[resolverMapping.type] || {} // temporary hack. This will be refactored as part of AEROGEAR-3436
         resolvers[resolverMapping.type][resolverMappingName] = buildPostgresResolver(
           client,
           resolverMapping.requestMapping,
           resolverMapping.responseMapping
         )
       } catch (ex) {
+        console.log(ex)
         console.log('Error while building Postgres resolver for mapping: ' + resolverMappingName)
         throw new Error('Error while building Postgres resolver for mapping: ' + resolverMappingName)
       }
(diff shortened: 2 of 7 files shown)


server/lib/resolverMaker.js
```diff
@@ -4,50 +4,52 @@ Handlebars.registerHelper('toJSON', function (json) {
...
-          resolvers[resolverMappingName][key] = buildPostgresResolver(
-            dataSourceClient,
-            value.requestMapping,
-            value.responseMapping
-          )
-        } catch (ex) {
-          console.log('Error while building Postgres resolver for mapping: ' + key)
-          throw new Error('Error while building Postgres resolver for mapping: ' + key)
-        }
+    if (['Query', 'Mutation'].indexOf(resolverMapping.type) < 0) {
```
david-martin: this conditional can be removed now.

darahayes: Hmm I tried to remove it but some tests failed as a result. They failed because the expected behaviour is to exit early if an Unknown query type is used e.g. if a user specifies a resolverMapping with `type: 'foo'` for example.

darahayes: I think it might be best to leave this in for now but also include `Subscriptions`

david-martin: I don't think we an put any restriction here as resolvers can be attached to any custom type as defined by the user in the GraphQL Schema.

See this example where a bunch of different resolvers are needed for parts of the schema besides Query, Mutation & Subscription
https://github.com/apollographql/starwars-server/blob/master/data/swapiSchema.js#L442-L542

david-martin: Another example.
For 1:M relationships defined in you GraphQL Schema, you'll need to attach a Resolver to the 1 side of that relationship e.g.

```graphql
type Post {
  text: String
  comments: [Comment]
}

type Comment {
  text: String
}
```

A resolver is needed on the `comments` field of the Post type e.g. 
```json
{
  "Post": {
    "comments": function() {/*...*/}
  }
}
```

aliok: During my work, I added test cases for the current state of the code, as _characterization tests_. I also added handling code for some base cases. I wasn't sure about everything in those base cases, but it is super easy to remove those anyway.

BTW, I like this: improving stuff one step at a time.

aliok:
Haven't verified, but it all looks good.

server/lib/resolverMaker.js
darahayes: @david-martin thanks for this great explanation! I will need to make some modifications to one or two tests in order for this to work.

README.md
```diff
@@ -85,3 +85,30 @@ The baseline architecture is shown below:
 6. Configured Authentication & Autohorizatin checks are applied
 7. Logging & Metrics data is gathered from the Server & connected Clients
 
+## Running Tests
+
+```
+npm run test
+```
+
+### Debugging Individual Tests
```
aliok: :+1:

PR #11, opened 2018-07-03
[WIP] feat: establish pattern for datasource and resolver creation

darahayes:
Related Ticket: https://issues.jboss.org/browse/AEROGEAR-3436

At a high level this PR makes it easy to add new data sources and new resolvers using something I'm calling **Data Source Parsers** and **Resolver Builders**

@david-martin @aliok I'd like to get your input before I do the final bits of tidy up + implementing tests.

### Verification

This PR is in a runnable state and it is possible to make queries against nedb. I've changed the `npm run dev` command to use the `resolver-mappings-nedb.example.json` file.

Start it up and try out the queries in the graphiql console.

### Summary

At a top level we still have the Schema Parser. The Schema Parser does two things:

* Parses the Data Sources (using lib/datasources/dataSourceParser.js)
* Builds the Resolvers (using lib/resolvers/resolverBuilder.js)

Now the dataSourceParser module makes use of individual parsers under the hood. There is a parser per type of datasource.

The resolverBuilder module makes use of individual resolver builders under the hood. Again, there is a resolver builder per type of datasource.

I have come up with an initial design for how a parser and a resolver builder should work. i.e. the inputs and expected outputs. You can read below how to add new ones.

## Adding New Data Sources

Firstly, take a look at data-sources.example.json

```json
{
  "p1": {
    "type": "postgres",
    "config": {
      "user": "postgres",
      "password": "mysecretpassword",
      "database": "postgres",
      "host": "127.0.0.1",
      "dialect": "postgres"
    }
  },
  "nedb_notes": {
    "type": "nedb",
    "config": {
      "collection": "notes"
    }
  }
}
```

I'm calling these **Data Source Definitions**. They must have a `type` and `config` field. To take a Data Source Definition and to turn it into a **Datasource** (the internal understanding of a datasource), the appropriate **parser** is used. You can see the parsers in `lib/datasources/parsers`.

Let's look at the nedb.js parser as an example:

```javascript
const Datastore = require('nedb')

const type = 'nedb'

function createNEDBDatasource (config = {}, connect = true) {
  return { client: new Datastore(), type } // this object is the datasource
}

// this object is the parser
module.exports = {
  createDataSource: createNEDBDatasource
}
```

A parser is an object with a function called `createDataSource(config, connect)`. This function must an object like so: `{ type: 'name of type', client: <data source client, e.g. a database connection>}`

New datasources can be supported by adding a new parser file under `lib/datasources/parsers`, and then exporting it from `lib/datasources/parsers.js`. This will all work so long as the parser follows the structure.

### Adding New Resolvers

The process for adding new resolvers is similar to adding new Data Sources.

Under `lib/resolvers/builders` we export the supported resolver builders.

A Resolver Builder is an object with a function called `buildResolver(dataSourceClient, compiledRequestMapping, compiledResponseMapping)`.

* `dataSourceClient` - something like a db connection or a nedb collection
* `compiledRequestMapping` - An already compiled handlebars template of the request mapping
* `compiledResponseMapping` - An already compiled handlebars template of the response mapping.

The builder must return a Graphql resolver function. Take a look at the postgres one for example:

```javascript
const JSONParse = require('json-parse-safe')

function buildPostgresResolver (dataSourceClient, compiledRequestMapping, compiledResponseMapping) {
  return function resolve (obj, args, context, info) {
      const queryString = compiledRequestMapping({
        context: {
          arguments: args
        }
      })

      dataSourceClient.query(queryString, [], (err, res) => {
        if (err) return reject(err)
        // TODO: should we end the connection with each request?
        // dataSourceClient.end()

        const responseString = compiledResponseMapping({
          context: {
            result: res.rows
          }
        })

        let { value, error } = JSONParse(responseString)

        if (error) {
          // TODO better error message back to user when this happens
          return reject(error)
        }

        return resolve(value)
      })
    })
  }
}

module.exports = {
  buildResolver: buildPostgresResolver
}
```

It is possible to add new resolver builders by creating a file that follows this structure and placing it inside `lib/resolvers/builders` and ensuring it is exported.

### DIFF
--- a/server/lib/resolvers/resolverMapper.js
+++ b/server/lib/resolvers/resolverMapper.js
@@ -34,25 +34,21 @@ module.exports = function (dataSources, resolverMappings) {
     let builder = resolverBuilders[type]
 
     if (builder) {
-      try {
-        const { compiledRequestMapping, compiledResponseMapping } = compileMappings(resolverMapping.requestMapping, resolverMapping.responseMapping)
-        if (builder.buildResolver && typeof builder.buildResolver === 'function') {
-          const resolver = builder.buildResolver(
-            client,
-            compiledRequestMapping,
-            compiledResponseMapping
-          )
+      const { compiledRequestMapping, compiledResponseMapping } = compileMappings(resolverMapping.requestMapping, resolverMapping.responseMapping)
+      if (builder.buildResolver && typeof builder.buildResolver === 'function') {
+        const resolver = builder.buildResolver(
+          client,
+          compiledRequestMapping,
+          compiledResponseMapping
+        )
 
-          resolvers[resolverMapping.type] = resolvers[resolverMapping.type] || {}
-          resolvers[resolverMapping.type][resolverMappingName] = resolver
-        } else {
-          throw new Error(`resolver builder for ${type} missing build resolver function`)
-        }
-      } catch (ex) {
-        console.log(ex)
-        console.log(`Error while building resolver of type ${type} for mapping: ${resolverMappingName}`)
-        throw new Error(`Error while building resolver of type ${type} for mapping: ${resolverMappingName}`)
+        resolvers[resolverMapping.type] = resolvers[resolverMapping.type] || {}
+        resolvers[resolverMapping.type][resolverMappingName] = resolver
+      } else {
+        throw new Error(`resolver builder for ${type} missing buildResolver function`)
       }
+    } else {
+      throw new Error(`No resolver builder for type: ${type}`)
     }
   })
 


darahayes:
@aliok @david-martin This PR includes quite a lot of changes to directory structure. It might be worth fetching this branch and reviewing locally.

server/lib/resolvers/resolverBuilder.js
```diff
@@ -0,0 +1,60 @@
+const _ = require('lodash')
+const resolverBuilders = require('./builders')
```
aliok: This file is already resolverBuilder.js, but in the file something else is named as resolverBuilder.

I would rename this file. Maybe to "resolverMapper" or something.

server/lib/resolvers/resolverBuilder.js
```diff
@@ -0,0 +1,60 @@
...
+          const resolver = builder.buildResolver(
+            client,
+            compiledRequestMapping,
+            compiledResponseMapping
+          )
+
+          resolvers[resolverMapping.type] = resolvers[resolverMapping.type] || {}
+          resolvers[resolverMapping.type][resolverMappingName] = resolver
+        } else {
+          throw new Error(`resolver builder for ${type} missing build resolver function`)
```
aliok: This `throw` doesn't make that much sense. Maybe do like:
```
if (builder) {
  try {
    const {compiledRequestMapping, compiledResponseMapping} = compileMappings(resolverMapping.requestMapping, resolverMapping.responseMapping)
  } catch (ex) {
     ...
  }
  if (builder.buildResolver && typeof builder.buildResolver === 'function') {
    try{
      const resolver = builder.buildResolver(
        client,
        compiledRequestMapping,
        compiledResponseMapping
      )

      resolvers[resolverMapping.type] = resolvers[resolverMapping.type] || {}
      resolvers[resolverMapping.type][resolverMappingName] = resolver
    } catch (ex) {
      console.log(ex)
      console.log(`Error while building resolver of type ${type} for mapping: ${resolverMappingName}`)
      throw new Error(`Error while building resolver of type ${type} for mapping: ${resolverMappingName}`)
    }
  } else {
    throw new Error(`resolver builder for ${type} missing build resolver function`)
  }
} else {
  throw Error ("No builder for this type")
}
```

aliok: My suggestion has too many if/else statements but you get the idea.

server/lib/resolvers/builders/nedb.js
```diff
@@ -0,0 +1,69 @@
...
+          dataSourceClient.remove(query, options, mapResponse)
+          break
+        default:
+          return reject(new Error(`Unknown/unsupported nedb operation "${operation}"`))
+      }
+
+      function mapResponse (err, res) {
+        if (err) return reject(err)
+
+        var responseString = compiledResponseMapping({
```
aliok: var --> let/const

darahayes:
@aliok thanks for your valuable feedback. I think I've covered all of your suggestions. It turns out that try catch block in `resolverMapper.js` was unnecessary because the builder functions simply return a closure. There is nothing that could `throw` in there.

PR #9, opened 2018-07-02
[WIP] feat: Read all graphql configuration from postgres

david-martin:
feat: Read Data Source, Schema & Resolver configuration from posgres

The `sequelize` module is used for this.
A `DataSource`, `GraphQLSchema` & `Resolver` model is defined.
On startup, the first schema is retrieved from the database,
and all Resolvers & DataSources are read and passed into the schema parser
logic.

The local dev setup has changed due to the introduction of postgres.
See README for details.
The gist of it is:

```
npm run db:start
npm run db:seed      # Creates sequelize model tables & seeds some example data
npm run dev
```

Access graphiql on http://localhost:8000/graphiql

to stop:

```
npm run db:stop
```

*Other changes in this PR*

* remove example json files as they are no longer used
* remove k8s setup from README (to be replaced by a link to APB repo/setup in future)
* a change to the resolverMaker & datasourceParser logic to look for a `field` & `name` key respectively on each item in the item array, rather than using the key of an object

TODO:

* [x] Add models for GraphQLSchema, Resolver & Data Sources, removing all file based config
* [x] Merge with in-memory poc changes
* [x] Verify schema is OK for the Admin UI
* [ ] Tests

Separate follow up PR for:

* Auto reload the schema if the database table changes (any creates/deletes/updates)

PR #12, opened 2018-07-04
feat: Read Data Source, Schema & Resolver configuration from posgres

darahayes:
**Note** This PR is the same as #9 but only the relevant changes are cherry picked. #9 was rebased on #11 (which wasn't merged at the time). When it got merged things got a little messed up. I've cherry picked @david-martin's commits to retain his authorship. All credit goes to him.

feat: Read Data Source, Schema & Resolver configuration from posgres

The `sequelize` module is used for this.
A `DataSource`, `GraphQLSchema` & `Resolver` model is defined.
On startup, the first schema is retrieved from the database,
and all Resolvers & DataSources are read and passed into the schema parser
logic.

The local dev setup has changed due to the introduction of postgres.
See README for details.
The gist of it is:

```
npm run db:start
npm run db:seed      # Creates sequelize model tables & seeds some example data
npm run dev
```

Access graphiql on http://localhost:8000/graphiql

to stop:

```
npm run db:stop
```

*Other changes in this PR*

* remove example json files as they are no longer used
* remove k8s setup from README (to be replaced by a link to APB repo/setup in future)
* a change to the resolverMaker & datasourceParser logic to look for a `field` & `name` key respectively on each item in the item array, rather than using the key of an object

TODO:

* [x] Add models for GraphQLSchema, Resolver & Data Sources, removing all file based config
* [x] Merge with in-memory poc changes
* [x] Verify schema is OK for the Admin UI
* [ ] Tests

Separate follow up PR for:

* Auto reload the schema if the database table changes (any creates/deletes/updates)


server/lib/resolvers/resolverMapper.test.js
```diff
@@ -11,30 +11,32 @@ const resolverMapper = require('./resolverMapper')
...
 
-  const resolverMappings = {
-    'q1': {
-      'type': 'Query',
-      'dataSource': 'p1',
-      'requestMapping': 'q1_requestMapping {{ var }}',
-      'responseMapping': 'q1_responseMapping {{ toJSON "foo" }}'
+  const resolverMappings = [{
+    'type': 'Query',
+    'field': 'q1',
```
aliok: @darahayes I wonder why Dave named this as 'field'? Isn't it basically the mappingKey or mappingId?

darahayes: @aliok I'm not really sure to be honest. It makes the code a bit harder to follow also. I would much prefer `mappingKey`, `mappingId` or even just `name`

darahayes: Do note that the database schema for resolvers mappings has an `id` column as well as a `field` column.

aliok:
Tested these with GraphiQL:
```
query q1 {
  listNotes {
    _id
    title
    content
    createdAt
    updatedAt
  }
}

query q2 {
  readNote(id: "mGuebHrroIYeqlzw") {
    _id
    title
    content
  }
}

mutation m1 {
  createNote(
    title: "Hello"
    content: "Hello World!"
  ) {
    title
    content
  }
}

mutation m2 {
  updateNote(_id: "mGuebHrroIYeqlzw", title: "Hi", content: "Hello again World!") {
    _id
    title
    content
    createdAt
    updatedAt
  }
}

mutation m3 {
  deleteNote(id: "abc") {
    _id
    title
    content
    createdAt
    updatedAt
  }
}
```
And saw some problems. Will comment inline.

examples/create_example_data.sql
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    # The mutation type, represents all updates we can make to our data
+    type Mutation {
+      createNote(
+        title: String,
+        content: String,
+      ): Note
+      updateNote(
+        _id: String,
```
aliok: As in other methods, argument name should be "id" not "_id"

examples/create_example_data.sql
```diff
@@ -0,0 +1,45 @@
...
+      content: String
+      createdAt: String
+      updatedAt: String
+    }
+
+    type Subscription {
+      noteCreated: Note
+    }','2018-07-03 09:49:57.062 +00:00','2018-07-03 09:49:57.062 +00:00') RETURNING *;
+
+INSERT INTO "Resolvers" ("id","type","field","requestMapping","responseMapping","createdAt","updatedAt","DataSourceId") VALUES (DEFAULT,'Query','readNote','{"operation": "findOne","query": {"_id": "{{context.arguments._id}}"}}','{{toJSON context.result}}','2018-07-03 10:11:30.054 +00:00','2018-07-03 10:11:30.054 +00:00','1') RETURNING *;
```
aliok: `{{ context.arguments._id }}` should be `{{ context.arguments.id }}` because the parameter name is `id`

aliok:
I pushed a new commit and now the queries/mutations above all work.
The only problem I notice is, the new document is not returned when using `updateNotes` mutation

darahayes:
@aliok That is unrelated to this PR though. I am happy to investigate. Most likely it's something to do with the actual resolver function that's built.

examples/create_example_data.sql
```diff
@@ -0,0 +1,45 @@
...
+
+INSERT INTO "GraphQLSchemas" ("id","schema","createdAt","updatedAt") VALUES (DEFAULT,'schema {
+      query: Query
+      mutation: Mutation
+      subscription: Subscription
+    }
+
+    # The query type, represents all of the entry points into our object graph
+    type Query {
+      readNote(id: String): Note
```
darahayes: @aliok I'm just wondering should we change this to `_id`. Same with `updateNote` and `deleteNote`. Having `_id` returned but needing to specify `id` for updates, reads, deletes is a bit weird. What do you think?

aliok: @darahayes I think it is better if we have the parameter name as `id`. I updated param names for update and delete to `id` in my commit as well.

aliok:
@darahayes yeah, let's not fix results of `updateNote` and `deleteNote` right now.

aliok:
Good to merge!

examples/create_example_data.sql
darahayes: @aliok sounds good to me!

aliok:
BTW, I saw Dave mentioned that the tests are missing, but I don't think it makes sense to create unit tests for the new stuff. Tests are adapted for the existing stuff.
We eventually need to come up with integration tests! @darahayes I will create a JIRA for integration tests

PR #13, opened 2018-07-04
[WIP] Hot schema reload

aliok:
Tasks:
* [X] Initial impl - using the pubsub library directly
* [x] Abstraction layer, instead of calling the lib directly in server.js
* [x] Decide on the "channel name"
* [x] Throttle/debounce the channel listener
* [x] Graceful start (service should still start if there is no schema available)
* [x] Unit tests

### Triggering a hot schema reload:
Create a new file in the root of the source tree and run it directly:
```
process.env.POSTGRES_USERNAME = 'postgresql'
process.env.POSTGRES_PASSWORD = 'postgres'
process.env.POSTGRES_DATABASE = 'aerogear_data_sync_db'
process.env.POSTGRES_HOST = '127.0.0.1'
process.env.POSTGRES_PORT = '5432'

const config = require('./server/config')
let {postgresConfig} = config
console.log(postgresConfig)

const PGPubsub = require('pg-pubsub')

var pubsubInstance = new PGPubsub({
  user: postgresConfig.username,
  host: postgresConfig.host,
  database: postgresConfig.database,
  password: postgresConfig.password,
  port: postgresConfig.port
})

pubsubInstance.publish('aerogear-data-sync-config', {})
```


aliok:
@darahayes I will push more stuff here.

aliok:
@darahayes this is now in a "reviewable state". I followed the design you had for the resolvers/parser and loaded whatever file in "listeners" directory.
We now have a new config option as env var `SCHEMA_LISTENER_CONFIG`. It accepts JSON:
```
{ 
  type: 'postgres',
  config: { 
    channel: 'aerogear-data-sync-schema',
     database: 'postgres',
     username: 'postgres',
     password: 'mysecretpassword',
     host: '127.0.0.1',
     port: '5432' 
  } 
}
```

server/config/index.js
```diff
@@ -14,6 +14,29 @@ var config = {
     password: process.env.POSTGRES_PASSWORD || 'postgres',
     host: process.env.POSTGRES_HOST || '127.0.0.1',
     port: process.env.POSTGRES_PORT || '5432'
+  },
+  schemaListenerConfig: undefined
+}
+
+if (process.env.SCHEMA_LISTENER_CONFIG) {
```
darahayes: I'm not a big fan of supplying a JSON string as an environment variable. I've had previous (ops) experience where this was the most difficult thing ever.
technologies like Jenkins, CircleCI, Kubernetes, docker-compose, bash etc. All of these handle and escape these strings slightly differently and in my experience, It's really hard to get this to work everywhere.

I know it seems hacky but base64 is super portable. If really you want to supply this config as a JSON string I would suggest we assume this to be base64 encoded and then decode it here as part of startup. It would look something like this:

```
Buffer.from(process.env.SCHEMA_LISTENER_CONFIG, 'base64').toString()
```

About the JSON parsing... I had a discussion about user supplied config when working on aerogear-app-metrics. It was to do with something trivial like setting log levels via environment variables. I originally wanted to start the app with reasonable defaults in a case where the user supplied broken config.

The conclusion we arrived at was if a user supplied broken configuration, then the best option is to just crash the service because that service might start behaving in a way the user didn't expect and they won't know why.

server/lib/schemaListeners/schemaListenerCreator.js
```diff
@@ -0,0 +1,18 @@
+const listeners = require('./listeners')
+
+module.exports = function (schemaListenerConfig, callback) {
+  for (let type of Object.keys(listeners)) {
```
darahayes: Oh man I love this syntax. I didn't know you could do this 😄

darahayes:
@aliok I've left some small comments but overall I really like the direction this is going in 👍

server/config/index.js
aliok: About JSON in env vars: you're right that it could cause problems. We need to change it.
I think base64 idea is good. But, one thing we need to keep in mind is, how easy it is to create a bas64 encoded string with those techs (Jenkins, CircleCI, Ansible etc). Let's think about this for a minute.

The other way, and crappier way, is to use multiple env vars like:
For pg node Postgres channels:
- SCHEMA_LISTENER_TYPE: 'postgres'
- POSTGRES_CHANNEL_NAME: 'aerogear-something'
For the imaginary message broker XYZ:
- SCHEMA_LISTENER_TYPE: 'xyz'
- XYZ_TOPIC_NAME: 'aerogear-something'
- XYZ_HOST: 'example.com'
- XYZ_USERNAME: 'user'
- ...

The reason I went with JSON config in env vars is I didn't like all those strings (env var keys) that are used in some cases and not used in other cases.
I just wanted to bring this way to discussion so that we can have answers when somebody complains about base64 :smiling_imp:

server/lib/schemaListeners/listeners/postgres.js
```diff
@@ -0,0 +1,17 @@
+const PGPubsub = require('pg-pubsub')
+
+module.exports = function (config, callback) {
```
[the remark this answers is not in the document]
aliok: +1

server/config/index.js
aliok: > The conclusion we arrived at was if a user supplied broken configuration, then the best option is to just crash the service because that service might start behaving in a way the user didn't expect and they won't know why.

This is a good point. Then I suggest that we get rid of all the defaults in the code and just have those values in the dev scripts. Then we can complain whenever something is missing. Purpose of defaults to me are to start things faster on devl anyway.

darahayes: Yup no config is fine, it usually means we're in dev or that the user is happy with the defaults. Broken config means the user messed up and we shouldn't start in an unknown state.

With regards to the question of how difficult is base64 encoding something? It's pretty simple anywhere with a bash shell.

```bash
echo 'Hello world' | base64
```

darahayes: I'm not a massive fan of the base64 thing but I've genuinely spent an entire week trying to escape a JSON string so a CI server could feed it into an app before and I don't ever want to do that again 😄 

I guess the reason you're not a fan of the other solution is because it's inherently tied to Postgres and it kinda breaks the abstraction layer we're trying to place above that.

To be honest, as long as the underlying code is abstracted well enough that we could swap it out in the future, I wouldn't mind the other approach because refactoring a config file is a minor thing.

I will defer this to you :)

server/lib/schemaListeners/schemaListenerCreator.js
```diff
@@ -6,8 +7,17 @@ module.exports = function (schemaListenerConfig, onReceive) {
...
-        return // return to end the loop
+        // "onReceive" will cause the server to reload the configuration which could be costly.
+        // don't allow doing it too often!
+        // we debounce the "onReceive" callback here to make sure it is debounced
+        // for all listener implementations.
+        // that means, the callback will be executed after the system waits until there
+        // is no request to call it for N milliseconds.
+        // like, when there's an evil client that notifies the listener every 100 ms,
+        // we still wait for N ms after the notifications are over
+        const debouncedOnReceive = _.debounce(onReceive, 500)
```
darahayes: Great idea!

server/server.js
```diff
@@ -42,17 +43,44 @@ async function buildSchema (models) {
...
 
   const resolvers = await models.Resolver.findAll({
     include: [models.DataSource]
   })
-  const resolversJson = resolvers.map((resolver) => {
+  let resolversJson = resolvers.map((resolver) => {
     return resolver.toJSON()
   })
 
+  if (_.isEmpty(graphQLSchemaString) || _.isEmpty(dataSourcesJson) || _.isEmpty(resolversJson)) {
```
darahayes: If you look at line 40 we define `let graphQLSchemaString = '{}'`

https://github.com/aerogear/data-sync-server/pull/13/files/7b540109760d1d7e1e57e2b1bb790f8b68f2a7f0..15cb27b90800f444fdfd18e47e9bba590f8bf466#diff-ea75b530d74f94670cea053feb8b961bR40

This means there is a small edge case where the data sources + resolvers are available but no schema and this condition would return truthy. It could lead to unexpected behaviour or crashing

darahayes:
@aliok once that minor little bug is sorted then I'm super happy with this PR. We've made fantastic progress!

server/server.js
aliok: Good catch! I've only tried with all tables empty and didn't catch that.

aliok:
@darahayes this PR is ready for the final review.
BTW, I converted the schema listener to a class (OR type OR constructor) so that we can keep the state and have a reference to `pubsubInstance`. This allowed separation of instantiating the listener and activating it (`start` method). This was necessary anyway for graceful shutdowns.

aliok:
@pb82 we now use base64 encoding for pubsub config. (see comment https://github.com/aerogear/data-sync-server/pull/13#discussion_r200173946)

PR #14, opened 2018-07-04
fix: local development environment improvements

darahayes:
This PR does the following:

* The most notable thing is the `Sequelize` folder which looks like this:

```
sequelize/
├── config
│   └── database.js
├── models
│   ├── dataSource.js
│   ├── graphQLSchema.js
│   ├── index.js
│   ├── resolver.js
│   └── user.js
├── seeders
│   └── 20180704100455-notes-example.js
└── sync_models.js
```

This is a sort of sequelize convention. Normally the `seeders`, `config` and `models` folder would be at a top level but I decided it would be nice if we could isolate all the sequelize stuff into its own folder. This was configured with the `.sequelizerc` file.

* I've removed the example `.sql` seed script and replaced it with a javascript file found under `sequelize/seeders`. This takes advantage of the seeding framework Sequelize has and lets us define seed data much easier. Sequelize has a migrations framework which works very similarly which we might take advantage of in the future.
* Using docker-compose to spin up a postgres instance

This doesn't make any change to the local development flow or to how the application works, but it's a bit more maintainable going forward.

One thing to note is that the `npm run db:init` command will drop the tables every time. This is very handy from a local development perspective where you might want to change and reload data very often.

## Verification

Follow the new instructions in the Readme and see if you can start up the database with docker-compose, initialize it and run the server. Then verify that everything works as expected.


aliok:
:eye:

sequelize/sync_models.js
```diff
@@ -0,0 +1,8 @@
+const postgresConfig = require('../server/config').postgresConfig
```
aliok: I know you wanted to have sequelize things isolated, but can we move this file to 'scripts' folder? This is not a library/module that's to be imported.

README.md
```diff
@@ -4,58 +4,48 @@ GraphQL based data sync server for mobile, with backend integration capabilities
...
-Go to http://localhost:8000/graphiql for an interactive query brower.
-The graphql endpoint is at `/graphql`.
-The subscriptions websocket is at `/subscriptions`.
-
-*Inspecting Postgres*
+Use docker compose to start the database.
 
 ```
-npm run db:shell
+docker-compose up db
```
aliok: `docker-compose up postgres` worked

README.md
```diff
@@ -4,58 +4,48 @@ GraphQL based data sync server for mobile, with backend integration capabilities
...
+npm run db:init
 ```
 
-## Running on Kubernetes
+`npm run db:init` sets up the necessary tables and seeds the database with data useful for local development. **It is a destructive action.** It drops and recreates the tables every time.
 
-TODO
 
-## Architecture
+### Start the Server
```
aliok: Can you add a "clean up" section like how to stop and remove Postgres container?

aliok:
@darahayes left some comments about minor things. verified the devl workflow, and made sure the example data is still the same. all good.

README.md
aliok: I am not talking about hitting Ctrl+C to stop the container. Talking about `docker-compose rm` to remove the stopped containers since when you do `docker-compose up`, existing stopped containers are started again (feature of docker-compose).

darahayes: Great idea!

README.md
darahayes: Ahh typo :)

sequelize/sync_models.js
darahayes: Good point. Come to think of it, I will also add some protection to this to ensure it cannot run with `require()`

darahayes:
One minor thing I've also done. I've removed `nodemon.json`. This was honestly making the `npm run dev` command unusable for me because nodemon would crash if there were unfixable lint issues even in files that you're not working on. This happens way too often IMO. The precommit hook and the CI server ensure that stuff is linted properly.

scripts/sync_models.js
```diff
@@ -0,0 +1,12 @@
...
+
+const forceDrop = process.env.FORCE_DROP === 'true'
+
+if (require.main === module) {
+  require('../sequelize/models')(postgresConfig).sequelize.sync({force: forceDrop}).then(() => {
+    console.log('sequelize done')
+    process.exit(0)
+  })
+} else {
+  throw Error('This file should not be imported. Ever.')
```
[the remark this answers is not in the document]
darahayes: Yeah I just remembered that weird `if __name__ == 'main'` thing from python and thought 'I wonder if this can be done in Node?' 😄

PR #15, opened 2018-07-05
fix: use dotenv to make testing different configs easy

darahayes:
The `dotenv` module is a very widely used module to do what the old `dev_env.sh` script did. Although the app has reasonable defaults for local dev, there may be a case where you want to try out different configs locally and typing out environment variables is a pain. Simply put the values in your `.env` file and try it out locally.

I've left the file empty because duplicating the defaults leads to confusion. (Hey I changed the code but for some reason it's still using the old config).

Additionally, I've added `.env` to `.dockerignore` to ensure it's never included in any images


aliok:
This is cool! But, this is only to be used during devl. So, why not moving the defaults from config.js to here? Defaults in config.js is also for devl only.

darahayes:
I guess I like having the actual defaults a feature of the code whereas `.env` is just a nicer syntax than manually typing `MYVAR_1=some MYVAR_2=additional MYVAR_3=options npm run dev`.

Take `config.server.port` or `config.postgresConfig.database` as examples. We'll more than likely use the default values provided in the code in live scenarios. But it would be a little annoying if we absolutely had to provide those values via env variables.

aliok:
Looks good, but let's merge https://github.com/aerogear/data-sync-server/pull/13 first and then this one.

PR #16, opened 2018-07-06
add health endpoint

pb82:
add health endpoint

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -45,6 +45,7 @@ module.exports = async ({ graphQLConfig, graphiqlConfig }, models) => {
 
   // TODO Move this to the Admin UI
   app.get('/graphiql', graphiqlExpress(graphiqlConfig))
+  app.get('/healthz', (req, res) => res.sendStatus(200))
 
   // Wrap the Express server
   const server = http.createServer(app)


aliok:
Looks good!

@pb82 when I think about this, we need to make sure that application exits gracefully when cannot connect to Postgres (2 cases: config reading, channel listening).
Otherwise, we might still get good health status with this new endpoint even though things are not working.

Could you test that quickly, if interested?

pb82:
@aliok sure, i'll add something like this to the UI too and then to this PR

darahayes:
I would do some small research into how Sequelize handles connection errors just to be sure we don't end up with something like a healthcheck that kills the server because the database was just coming online, or because the connection temporarily dropped.

pb82:
@darahayes @aliok Sequelize suggest to use the `authenticate` method to test the connection: http://docs.sequelizejs.com/manual/installation/getting-started.html#test-the-connection

aliok:
> I would do some small research into how Sequelize handles connection errors just to be sure we don't end up with something like a healthcheck that kills the server because the database was just coming online, or because the connection temporarily dropped.

That's a valid point! I have very little experience how this is handled in other projects. cc @maleck13

pb82:
Closing in favor of new PR

PR #18, opened 2018-07-06
add health endpoint and database check

pb82:
add health endpoint (`/health`) and database check.

a successful check returns:

```
{"status":"ok","summary":"No issues to report. All tests passed without error.","details":[{"description":"Database connectivity","test_status":"ok","result":"Database connectivity check successful","runtime":8}]}
```

a failed check with stopped database returns:

```
{"status":"crit","summary":"A critical test item encountered an error. Please investigate this. See the \"details\" object for specifics.","details":[{"description":"Database connectivity","test_status":"crit","result":{"name":"SequelizeConnectionRefusedError","parent":{"errno":"ECONNREFUSED","code":"ECONNREFUSED","syscall":"connect","address":"127.0.0.1","port":5432},"original":{"errno":"ECONNREFUSED","code":"ECONNREFUSED","syscall":"connect","address":"127.0.0.1","port":5432}},"runtime":4}]}
```

Currently the status code even on a failed response is 200. Should we change that to 500?
*UPDATE*: returns status 500 on failed checks now (plus json payload)


aliok:
:eyes:

aliok:
Verified locally. All good!

server/health/healthchecks.js
```diff
@@ -0,0 +1,25 @@
+const postgresConfig = require('../config').postgresConfig
+const {addCriticalTest, runTests} = require('fh-health')
```
darahayes: Do we really need this library? Looked through the library and it's good but it just seems unnecessary

A much simpler and similar alternative would be something like this:

```javascript
// health/index.js

module.exports = function Health(db) {
  
  function databaseConnectivityCheck() {
      db.sequelize.authenticate()
        .then(resolve)
        .catch((err) => {
          console.error(err) // log the actual error
          reject('Database Connection Error') // return a generic one
        })
    })
  }

  function runChecks() {
    const checks = [ databaseConnectivityCheck() ] // can add more if we need
    return Promise.all(checks)
  }

  return runChecks
}
```

and then in `server.js`

```javascript
const runHealthChecks = require('./lib/health')(models)

app.get('/healthz', async (req, res) => {
  try {
    await runHealthChecks()
    res.json({ok: true})
  } catch(error) {
    res.status(503)
    res.json({ error })
  }
})
```

I feel like this is a simpler way to go but still allows us to achieve a similar thing as `fh-health.`

pb82:
@darahayes @aliok I rewrote the health checks to not use fh-health. The libary has a few other problems like returning the result as a string instead of an object wich required us to use `JSON.parse` and other things.

This implementation doesn't use any library, allows us to define multiple tests and returns a format that allows for labelling the various health checks without revealing anything sensitive.

A failed result looks like:
```
{"ok":false,"checks":[{"Database Connectivity":false}]}
```
(where `ok` is the overall result and `checks` contains the labelled individual results)

PR #20, opened 2018-07-11
Devl workflow for memeolist

aliok:
@darahayes 
This PR adds 2 additional npm commands.

* db:init:memeo - Creates the DB with memeo list schema and resolvers
* dev:memeo - Feeds the memeolist example queries to Graphiql so that when you go to localhost:8000/graphiql, you can see those queries.

The schema itself is not complete. I am working on it at the moment.

### DIFF
--- a/sequelize/models/resolver.js
+++ b/sequelize/models/resolver.js
@@ -2,8 +2,8 @@ module.exports = (sequelize, DataTypes) => {
   const Resolver = sequelize.define('Resolver', {
     type: DataTypes.STRING,
     field: DataTypes.STRING,
-    requestMapping: DataTypes.STRING,
-    responseMapping: DataTypes.STRING
+    requestMapping: DataTypes.TEXT,
+    responseMapping: DataTypes.TEXT
   })
 
   Resolver.associate = (models) => {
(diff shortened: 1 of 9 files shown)


sequelize/models/resolver.js
```diff
@@ -2,8 +2,8 @@ module.exports = (sequelize, DataTypes) => {
   const Resolver = sequelize.define('Resolver', {
     type: DataTypes.STRING,
     field: DataTypes.STRING,
-    requestMapping: DataTypes.STRING,
-    responseMapping: DataTypes.STRING
+    requestMapping: DataTypes.TEXT,
+    responseMapping: DataTypes.TEXT
```
aliok: These 2 were causing problems. STRING by default has a capacity of 255 chars in POSTGRES.
Using TEXT means unlimited capacity.

server/config/index.js
```diff
@@ -1,5 +1,13 @@
 require('dotenv').config()
 
+const fs = require('fs')
+
+let graphiqlQueryFileContent = ''
+
+if (process.env.GRAPHIQL_QUERY_FILE) {
+  graphiqlQueryFileContent = fs.readFileSync(process.env.GRAPHIQL_QUERY_FILE).toString()
```
darahayes: two minor nits:

* you can pass `'utf8'` as a second argument to `fs.readFileSync` and a string will be returned (as opposed to a buffer)
* perhaps we should surround this in try/catch to avoid crashing when a file doesn't exist

aliok:
@darahayes problems in comments are fixed. please have another look

PR #22, opened 2018-07-12
Memeolist proper

aliok:
This PR is about trying out the Memeolist schema on our sync server as per https://issues.jboss.org/browse/AEROGEAR-3441

This is the schema defined in the proposal: https://github.com/secondsun/proposals/blob/eb4bc07e472f9a96ee1d69c491f772ecb484eabb/dogfood.md#sample-schema

Trying out a real world example as a user helps with finding bugs and limitations.
I didn't just put the whole schema and then try to write resolvers. I am going step by step.

Here are the issues found so far:
* NeDB uses `_id` as the id field name. This causes portability issues such as a schema itself needs to be updated when switching the data source. Ideally, only the resolvers should've changed, not the schema.
As a solution to that, I implemented a 'convertNeDBIds' Handlebars helper that would convert `_id`s to `id`s. With this change, types in schemas can still use `id` as the id field.

* NeDB doesn't have an option to return the deleted docs. It only returns the number of deleted docs. So, the delete query definitions should be changed to return a number (num deleted) or boolean (if something is deleted).
I have also implemented a `toBoolean` helper to use in resolvers that could be useful in this case.

* In NeDB unlike queries and inserts, update operation's callback have a different signature. Needed to adapt the NeDB resolver builder.

* NeDB supports unique indices, but they need to be created in advance. We are not planning to support this with in mem data source. This should be documented.

Tasks:
- [X] allProfiles (query) : simple query
- [X] profile (query) : another simple query, but make sure the return result is proper when no profile found
- [X] createProfile
- [X] updateProfile
- [X] deleteProfile
- [ ] postMeme : Need to use NeDB's `$addToSet` or `$push` to insert stuff into `profile` doc's `memes` property
- [ ] allMemes: Get all `profile.memes` and order them by create/update time
- [ ] deleteMeme: Need to use  NeDb's `$pull`
- [ ] postComment: Insert a comment into `profile.memes[].comments[]`
- [ ] deleteComment
- [ ] What about the timestamps like `meme.createdOn`? Make it working properly with GraphQL and NeDB.
- [ ] Graphql returns an error instead of null when an object not found: https://stackoverflow.com/questions/51287139/graphql-return-a-type-with-non-nullable-id-field-as-a-query-result/51293035#51293035

**NOTE: We don't have to use the schema in the proposal as-is. For example, that schema doesn't accept a profile id in the `postMeme` mutation because it has the assumption that there's an auth in place. We should just add the profile id param to `postMeme`**

### DIFF
--- a/sequelize/seeders/20180711143600-memeolist-example.js
+++ b/sequelize/seeders/20180711143600-memeolist-example.js
@@ -28,17 +28,21 @@ const notesSchema = {
   
   type Meme {
     id: ID! @isUnique
+    photoUrl: String!
+    ownerId: String!
   }
   
   type Query {
     allProfiles:[Profile!]!
     profile(email: String!):Profile
+    allMemes:[Meme!]!
   }
   
   type Mutation {
     createProfile(email: String!, displayName: String!, biography: String!, avatarUrl: String!):Profile!
     updateProfile(id: ID!, email: String!, displayName: String!, biography: String!, avatarUrl: String!):Profile
     deleteProfile(id: ID!):Boolean!
+    createMeme(ownerId: String! photoUrl: String!):Meme!
   }
   
   type Subscription {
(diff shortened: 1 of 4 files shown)


wtrocki:
@aliok If we looking to match functionalities of the app then we do not need as much mutations. 

Currently we only need:
- postMeme
- allMemes

In future we may look into:
- getUserProfileByEmail
- postComment

wtrocki:
@aliok This could be useful: https://github.com/Glavin001/graphql-sequelize-crud

sequelize/seeders/20180711143600-memeolist-example.js
```diff
@@ -17,20 +17,30 @@ const notesSchema = {
   name: 'default',
   schema: `
   
-  schema {
-    query: Query
-    mutation: Mutation
-    subscription: Subscription
+  type Profile {
+    id: ID! @isUnique
```
wtrocki: Usually generated code contains fragement. Pattern is that frament is being used to defined type. Then it's also referenced in every crud operation. Schema defined this way is easier to maintain

aliok:
@wtrocki 

>@aliok If we looking to match functionalities of the app then we do not need as much mutations.

This PR is not about the current state of the memeolist app. We're just exploring the data sync service for identifying the problems, limitations and bugs.

wtrocki:
@aliok After reviewing existing code I think it's nice to get some of the stuff you have there into schema. It's really valuable PR from my point of view. 
@darahayes WDYT?

darahayes:
I wholeheartedly agree it would be good to get this merged. The current schema is becoming a little too basic for manual + e2e testing of features.

wtrocki:
@aliok Yes. I will make all required changes to Android app

darahayes:
@wtrocki go for it!

wtrocki:
Closing PR as I'm going to create new one with support for fragments etc.

PR #23, opened 2018-07-12
hack: allow db:init:memeo run inside container

darahayes:
This PR makes some changes such that it's possible to run `npm run db:init:memeo' inside the running container image.

This allows us to do a little hack where we can use the OpenShift UI to get a shell inside the running container and then we can manually execute `npm run db:init:memeo`. This will enable us to do the demo.


aliok:
Good to merge!

PR #25, opened 2018-07-13
Multiple schemas - adapt to DB model but not support it

aliok:
This PR adapts the sync service to the changed DB model which allows multiple schemas and which assigns a resolver to a schema.

However, we don't support multiple schemas at the moment. So, the process will exit when there are multiple.

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -4,9 +4,9 @@ const express = require('express')
 const bodyParser = require('body-parser')
 const {graphqlExpress, graphiqlExpress} = require('apollo-server-express')
 const cors = require('cors')
-const { log } = require('./lib/util/logger')
-const expressPino = require('express-pino-logger')({ logger: log })
-const { runHealthChecks } = require('./health')
+const {log} = require('./lib/util/logger')
+const expressPino = require('express-pino-logger')({logger: log})
+const {runHealthChecks} = require('./health')
 
 const schemaParser = require('./lib/schemaParser')
 const schemaListenerCreator = require('./lib/schemaListeners/schemaListenerCreator')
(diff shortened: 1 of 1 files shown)


server/server.js
```diff
@@ -92,10 +92,15 @@ module.exports = async ({graphQLConfig, graphiqlConfig, postgresConfig, schemaLi
 }
 
 async function buildSchema (models) {
-  const graphQLSchema = await models.GraphQLSchema.findOne()
+  const graphQLSchemas = await models.GraphQLSchema.findAll()
   let graphQLSchemaString = null
-  if (graphQLSchema != null) {
-    graphQLSchemaString = graphQLSchema.schema
+
+  if (!_.isEmpty(graphQLSchemas)) {
```
darahayes: What about only loading the schema with the name `default`? @pb82 how is this handled in the UI?

pb82: @darahayes Yes, thats what the UI will do: https://github.com/aerogear/data-sync-ui/pull/31/files#diff-ca481b05655823ba99892d2d34415a03R140

aliok:
I actually prefer making things fail when there's something we cannot handle. Otherwise, we need to document this and that.

But if you both think we should not fail and just read the schema with name 'default' I am ok with that too.

Let's vote.

thumbs up on this comment: fail
thumbs down on this comment: no fail, use 'default'

darahayes:
@aliok I agree we should make things fail when there's something we cannot handle.

However, I think that the existence of multiple schemas in the database is something we should be able to handle right now. (even if we don't do anything about it)

What I'm suggesting is for now is one of these two:

* we don't even bother looking for other schemas, we simply look for `default` and we fail if that isn't found.
* we load in all schemas, only use `default` and if that's not found then we fail.

aliok:
@darahayes please have another look

PR #26, opened 2018-07-16
fix: ensure handlebars does not escape unicode characters

darahayes:
I found an issue where the handlebars templates (request and response mappings) would escape unicode characters. For example, in the memeolist app this led to the `photoUrl` property being broken and it meant that in the app, the image could not be displayed

ping @aliok @wtrocki

### DIFF
--- a/server/lib/resolvers/compiler.js
+++ b/server/lib/resolvers/compiler.js
@@ -25,6 +25,7 @@ Handlebars.registerHelper('convertNeDBIds', function (json) {
 
 function compileMappings (requestMapping, responseMapping) {
   // use Handlebars.precompile to fail early during initialization
+  const noEscape = true
   try {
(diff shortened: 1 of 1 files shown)


aliok:
I wonder if this causes some kind of injection problems.

darahayes:
@aliok I'm not 100% sure. Handlebars's actual use case is as a templating engine for HTML documents. This escaping is there to escape stuff that is being injected into a HTML page. This is to avoid stuff like cross site scripting etc. Not sure how this affects us on the server side.

aliok:
@darahayes you're right! feel free to merge

aliok:
@wtrocki 

> This change alters handlebars so {{{ and {{ behave the same way

Sorry, couldn't understand if this is a problem or is it a feature?

PR #32, opened 2018-07-23
feat: initial subscriptions implementation

darahayes:
#### To do

* Tests
* Cleanup???

This is runnable right now though. You can try it out as follows:

* `npm run db:init:memeo`
* `npm run dev`
* Open `http://localhost:8000/graphiql` in two different tabs
* Leave one tab as it is with the sample queries
* In the other tab remove the sample queries and paste the following:

```
subscription memeAdded{
  memeAdded{
    id
    photoUrl
  }
}
```
* Run the subscription query and you should see the following message: `"Your subscription data will appear here after server publication!"`
* Run the createMeme query in the other window.

### How does the subscription piece work?

On the server side we define a new `memeAdded` subscription.

```
type Subscription {
    memeAdded(photoUrl: String):Meme!
}
```

Clients will be able to use this query to listen for updates. The photoUrl variable is optional (more on that later).

The next thing we need to be able to do is define a way to publish an event during some query or mutation. GraphQL uses a Pub/Sub model to do this. So when a query/mutation occurs, we can publish some events+data to a Pub/Sub mechanism and we can react to these events elsewhere. Below shows how a user could configure a `createMeme` resolver to also publish an event.

```js
{
    type: 'Mutation',
    field: 'createMeme',
    requestMapping: `{
      "operation": "insert",
      "doc": {
        "_type":"meme",
        "photoUrl": "{{context.arguments.photoUrl}}"
      }
    }`,
    responseMapping: '{{ toJSON (convertNeDBIds context.result) }}',
    publish: JSON.stringify({
      topic: 'memeCreated',
      payload: `{
        "memeAdded": {{ toJSON context.result }},
      }`
    }),
  }
```

The last piece of the puzzle is to be able to listen for those events and push them back down to a client. We can do this with the new `Subscription` config object. Take a look at the example one for the memeolist (included in this PR)

```js
{
    type: 'Subscription',
    field: 'memeAdded',
    topic: 'memeCreated',
}
```

This config tells our server to listen for messages published to the 'memeCreated' topic on the internal pubsub mechanism and to fire the `memeAdded` resolver as a result.

## Trying out Filtering

Filtering is an important aspect of Grapqhl subscriptions. In most real world apps, it doesn't make sense for a client to be listening for notifications about every single record that was added to a collection. It would be like getting a notification in slack every time anybody in the world sent a message.

Taking the Slack example further, on the client side we might subscribe to a messageAdded query as follows:

```
subscription messageAdded{
  messageAdded(channelId: '#random'){
    id
	channelId
    text
	userId
  }
}
```

The idea is we only want to listen for notifications in a particular slack channel - `#random`.

### How would this look on the server side?

How can we enable an end user of our service to define some kind of filtering that's relatively good?

I have come up with small DSL that allows the user to define some conditions in a **declarative** way using **JSON**. We add this filter to the **subscription** config object.

Here's when it might look like in the Slack example

```js
{
    type: 'Subscription',
    field: 'messageAdded',
    topic: 'messageCreated',
	filter: JSON.stringify({
      'eq': ['$payload.messageAdded.channelId', '$variables.channelId']
    })
  }
```

This filter would check that the channelId in the newly created message is equal to the channelId that the client passed as a variable when they subscribed. i.e. the user can say **give me real time updates about messages for this particular channel**.

The filter expression has access to a context object that looks like this: `{ payload: {...}, variables: {}}` where `payload` is the message received from the pubsub layer and variables contains whatever variables were passed by the client to initialize the subscription. Using the dollar syntax like `$payload.messageAdded.id` we can reference the context object.

I figured that 99% of the time, the basic use case will like the one above where we check that two things are equal. I made a more concise syntax for this particular case. The same example above can be expressed as follows:

```js
{
    type: 'Subscription',
    field: 'messageAdded',
    topic: 'messageCreated',
	filter: JSON.stringify({
      '$payload.messageAdded.channelId': '$variables.channelId'
    })
  }
```

This is the most basic usage of the filter language. You can do many more things. Try some of these out with the memeolist schema:

(you can modify the subscription object in the file located under `seeders`, then run the **db init and start commands again**)

```js
// subscribe only to new memes where payload.photoURL === google.com
{
    type: 'Subscription',
    field: 'memeAdded',
    GraphQLSchemaId: 2,
    topic: 'memeCreated',
    createdAt: time,
    updatedAt: time,
	filter: JSON.stringify({
      'eq': ['$payload.memeAdded.photoUrl', 'https://google.com']
    })
  }
```

```js
// subscribe only to new memes where payload.photoURL !== google.com
{
    type: 'Subscription',
    field: 'memeAdded',
    GraphQLSchemaId: 2,
    topic: 'memeCreated',
    createdAt: time,
    updatedAt: time,
	filter: JSON.stringify({
      '!eq': ['$payload.memeAdded.photoUrl', 'https://google.com']
    })
  }
```

```js
// subscribe only to new memes where payload.photoURL matches a regex
{
    type: 'Subscription',
    field: 'memeAdded',
    GraphQLSchemaId: 2,
    topic: 'memeCreated',
    createdAt: time,
    updatedAt: time,
	filter: JSON.stringify({
      'match': ['$payload.memeAdded.photoUrl', 'https:\/\/.*']
    })
  }
```

```js
// subscribe only to new memes where payload.photoUrl === https://google.com || payload.photoUrl == $variables.photoUrl
{
    type: 'Subscription',
    field: 'memeAdded',
    GraphQLSchemaId: 2,
    topic: 'memeCreated',
    createdAt: time,
    updatedAt: time,
	filter: JSON.stringify({
      'or': [
        { 'eq': ['$payload.memeAdded.photoUrl', 'https://google.com'] },
        { 'eq': ['$payload.memeAdded.photoUrl', '$variables.photoUrl']}
       ]
    })
  }
```

```js
// subscribe to new memes where payload.photoUrl === variables.photoUrl || (payload.photoUrl === https://facebook.com && payload.photoUrl === Dara)
{
    type: 'Subscription',
    field: 'memeAdded',
    GraphQLSchemaId: 2,
    topic: 'memeCreated',
    createdAt: time,
    updatedAt: time,
	filter: JSON.stringify({
      'or': [
        { 'eq': ['$payload.memeAdded.photoUrl', '$variables.photoUrl']},
        { 'and': [
                  { 'eq': ['$payload.memeAdded.photoUrl', 'https://facebook.com']},
                  { 'eq': ['$payload.memeAdded.owner', 'Dara'] }
                ]
       ]
    })
  }
```

### Why Did I do it this way?

* JSON format is **super portable**, for example - instead of defining the filter expression in the server config, a client app could define whatever filter they want and use it to listen for updates, **and it would be completely safe** because there's no way for them to inject code. This has the potential for some very useful applications IMO.
* I considered using handlebars to do something similar but it's very hacky because it doesn't actually evaluate things to `true` or `false`. It would also mean users would have to write a horrible syntax. In my opinion, this syntax is pretty intuitive.
* Inspired by MongoDB query language which I thought was very intuitive
* Flexibility, the code that evaluates these expressions could be extended in the future very easily.


server/lib/util/filterEvaluator.js
```diff
@@ -0,0 +1,91 @@
+const _ = require('lodash')
+
+module.exports = function ExpressionParser () {
```
darahayes: This is actually really generic and could be published as its own npm module with docs that demonstrate usage.

sequelize/seeders/20180711143600-memeolist-example.js
```diff
@@ -32,14 +32,25 @@ const notesSchema = {
   }
   
   type Subscription {
-    _: Boolean
+    memeAdded(photoUrl: String):Meme!
```
wtrocki: [Outside scope of this Pr] I think it will be more benefitial to add more fields to the meme like votes as this will help us to test various use cases.

server/server.js
```diff
@@ -112,6 +114,17 @@ module.exports = async ({graphQLConfig, graphiqlConfig, postgresConfig, schemaLi
   process.on('SIGQUIT', stopHandler)
   process.on('SIGINT', stopHandler)
 
+  // Set up the WebSocket for handling GraphQL subscriptions
+  /* eslint-disable no-new */
+  new SubscriptionServer({
```
aliok: this won't pick up new schema with hot reload

server/server.js
```diff
@@ -133,10 +146,8 @@ async function buildSchema (models) {
     }
   }
 
-  const dataSources = await models.DataSource.findAll()
-  let dataSourcesJson = dataSources.map((dataSource) => {
-    return dataSource.toJSON()
-  })
+  let dataSourcesJson = await models.DataSource.findAll({raw: true})
```
aliok: nice!

server/lib/schemaParser.js
```diff
@@ -2,10 +2,18 @@ const graphqlTools = require('graphql-tools')
 
 const dataSourceParser = require('./datasources/dataSourceParser')
 const resolverMapper = require('./resolvers/resolverMapper')
+const subscriptionsMapper = require('./subscriptions/subscriptionMapper')
+
+const PubSub = require('./pubsubNotifiers').InMemory
+
+module.exports = function (schemaString, dataSourcesJson, resolverMappingsJson, subscriptionMappingsJson) {
+  const pubsub = PubSub()
```
aliok: what's the plan to abstract this away?

server/lib/util/filterEvaluator.js
```diff
@@ -0,0 +1,91 @@
...
+    const arg1 = evalString(args[0], context)
+    const arg2 = evalString(args[1], context)
+    return expressionFn(arg1, arg2)
+  }
+
+  function evalString (str, context) {
+    if (typeof str !== 'string' || str.length <= 1) {
+      return str
+    }
+    var c1 = str[0]
```
aliok: I see some `var`s in the file

server/lib/resolvers/resolverMapper.js
```diff
@@ -53,3 +72,36 @@ module.exports = function (dataSources, resolverMappings) {
 
   return resolvers
 }
+
+function resolveAndPublish (resolverFn, publishOpts) {
+  return (obj, args, context, info) => {
+    return new Promise(async (resolve, reject) => {
+      try {
+        const result = await resolverFn(obj, args, context, info)
+        resolve(result)
```
aliok: +1 on resolving and then publishing

server/lib/resolvers/resolverMapper.js
```diff
@@ -53,3 +72,36 @@ module.exports = function (dataSources, resolverMappings) {
...
+        let compileOpts = {
+          context: {
+            result: result
+          }
+        }
+
+        let payload = compiledPayload(compileOpts)
+
+        // The InMemory pubsub implementation wants an object
+        // Whereas the postgres one would expect a string
```
aliok: as you noted, we need abstraction for this. let's not forget about that.

server/lib/util/filterEvaluator.js
aliok: I think it might be a better idea to extract this.

BTW, have you searched for a lib that does the same thing?

sequelize/seeders/20180711143600-memeolist-example.js
```diff
@@ -64,6 +75,12 @@ const resolvers = [
       }
     }`,
     responseMapping: '{{ toJSON (convertNeDBIds context.result) }}',
+    publish: JSON.stringify({
+      topic: 'memeCreated',
+      payload: `{
+        "memeAdded": {{ toJSON context.result }},
+      }`
+    }),
```
aliok: can you also add sample into `memeolist.query.graphql` file so that we don't have to type anything when we open Graphiql?

aliok:
@darahayes really great work! and, thanks for the long PR description which explains everything!
left some comments, but won't affect the general structure.

I think we would need some integration tests where we create 2 clients and mutate on one and listen on other one. Feel free to implement it as part of this pr OR we can implement it in a separate ticket/pr.

server/server.js
darahayes: 👍 This completely slipped my mind

server/lib/util/filterEvaluator.js
darahayes: I tried to search for something like this but I couldn't find anything. What would you even call this? I don't even know how to describe it in 2-3 words so I found it very hard to search for.

I think it would be great to extract. We should be constantly looking for opportunities to extract and open source internal libs like this. Where should the repo go? under github.com/aerogear?

server/lib/subscriptions/subscriptionMapper.js
```diff
@@ -0,0 +1,44 @@
+const _ = require('lodash')
+const { withFilter } = require('graphql-subscriptions')
+const { evalWithContext } = require('../util/filterEvaluator')()
+const { log } = require('../util/logger')
+
+module.exports = function mapSubscriptions (subsciptionMappings, pubsub) {
```
wtrocki: I would add some comments on what this module is doing.

server/lib/resolvers/resolverMapper.js
```diff
@@ -53,3 +72,36 @@ module.exports = function (dataSources, resolverMappings) {
...
 }
+
+function resolveAndPublish (resolverFn, publishOpts) {
+  return (obj, args, context, info) => {
+    return new Promise(async (resolve, reject) => {
+      try {
+        const result = await resolverFn(obj, args, context, info)
+        resolve(result)
+
+        const { pubsub, topic, compiledPayload } = publishOpts
```
wtrocki: I would wrap the code from here into separate function like `notifySubscriptions()` that could be used in different context.

wtrocki:
@darahayes Performing verification and adjusting memelist app

server/lib/subscriptions/subscriptionMapper.js
darahayes: Great suggestion

wtrocki:
@darahayes Adjusting now the memolist app for testing. Initial subscription looked like this. Do we support action filter from mobile. Something like here: https://github.com/aerogear/aerogear-android-example-apps/blob/07fcffdb17b4a7bd2543097b1b5e8a6aef159dc5/memeolist/app/src/main/graphql/org/aerogear/android/app/memeolist/graphql/memes.graphql#L17-L19

darahayes:
@wtrocki I have never seen this kind of syntax before. We definitely don't support it right now. Can you link to some Apollo/Gprahql documentation that shows this off in more detail? I'd like to learn more

aliok:
@darahayes @wtrocki that I think is from the schema generated by Graphcool. When you define a type it generates queries/mutations/subscriptions.
Example type:
```
type Note {
  id: ID! @isUnique
  text: String
}
```

Portion of generated stuff:
```
allNotes(filter: NoteFilter, orderBy: NoteOrderBy, skip: Int, after: String, before: String, first: Int, last: Int): NoteConnection!
...
input NoteFilter {
  # Logical AND on all given filters.
  AND: [NoteFilter!]

  # Logical OR on all given filters.
  OR: [NoteFilter!]
  id: ID

  # All values that are not equal to given value.
  id_not: ID

  # All values that are contained in given list.
  id_in: [ID!]

  # All values that are not contained in given list.
  id_not_in: [ID!]
...
}
```

That's still valid Graphql.

wtrocki:
Verified with graphql. Verified on the Android application.
Rebase needed! I will be nice to put some of the description from this PR into some rough docs in the repository so we will have some documentation draft.

darahayes:
@aliok Thanks for the great feedback. I believe I have incorporated all of it except the thing about extracting the filterEvaluator library. That will require a lot of extra time.

Let me know what you think!

server/server.js
```diff
@@ -53,6 +56,7 @@ module.exports = async ({graphQLConfig, graphiqlConfig, postgresConfig, schemaLi
       log.info('Received schema change notification. Rebuilding it')
       let newSchema
       try {
+        newSubScriptionServer(server, schema)
```
wtrocki: Should this be on the new schema?

server/lib/pubsubNotifiers/pubsubNotifier.js
```diff
@@ -0,0 +1,25 @@
...
+
+  if (!PubSubClass) {
+    throw new Error(`Unhandled pubsub type: ${pubsubConfig.type}`)
+  }
+
+  if (typeof PubSubClass !== 'function') {
+    throw new Error(`PubSub implementation for ${pubsubConfig.type} is missing a constructor`)
+  }
+
+  const pubsub = new PubSubClass(pubsubConfig.config)
```
camilamacedo86: Is not better add all const on top of it? All together.

server/lib/pubsubNotifiers/pubsubNotifier.js
```diff
@@ -0,0 +1,25 @@
+const notifiers = require('./notifiers')
+
+module.exports = function NewPubSub (pubsubConfig) {
+  const PubSubClass = notifiers[pubsubConfig.type]
+
+  if (!PubSubClass) {
```
camilamacedo86: Is not better encapsulate these if in functions the call them, something like as:
```javascript

module.exports = function NewPubSub (pubsubConfig) {
   const PubSubClass = notifiers[pubsubConfig.type]
  
   try {
          isValidType(PubSubClass);
          isFunction(PubSubClass);
          return getPublish(PubSubClass);
     } catch Error{
          throws Error;
    }

}

```

darahayes: Hi @camilamacedo86 thanks for reviewing! Your suggestion is very good and fair, but I do not see the need to move any of this stuff into separate functions right now because because this is the only place we will ever run these checks.

server/lib/pubsubNotifiers/pubsubNotifier.js
darahayes: @camilamacedo86 not sure what you mean, it wouldn't be possible to declare this at an earlier point in the code.

server/config/index.js
```diff
@@ -15,16 +17,19 @@ if (process.env.GRAPHIQL_QUERY_FILE) {
   }
 }
 
+const port = process.env.HTTP_PORT || '8000'
```
camilamacedo86: Add all in the top without space.

server/config/index.js
```diff
@@ -15,16 +17,19 @@ if (process.env.GRAPHIQL_QUERY_FILE) {
...
     tracing: true
   },
   graphiqlConfig: {
     endpointURL: '/graphql', // if you want GraphiQL enabled
-    query: graphiqlQueryFileContent
+    query: graphiqlQueryFileContent,
+    subscriptionsEndpoint: `ws://${hostname()}:${port}/subscriptions`
   },
   postgresConfig: {
     database: process.env.POSTGRES_DATABASE || 'aerogear_data_sync_db',
```
camilamacedo86: IHMO the fixed values as `'aerogear_data_sync_db'` could be declared as const on to of it.

server/lib/pubsubNotifiers/notifiers/inMemory.js
```diff
@@ -0,0 +1,25 @@
+const PubSub = require('graphql-subscriptions').PubSub
+const { log } = require('../../util/logger')
+
+function InMemoryPubSub (config) {
+  const client = new PubSub()
+
+  this.publish = function publish ({ topic, compiledPayload }, context) {
+    let payload = compiledPayload(context)
+    // The InMemory pubsub implementation wants an object
```
camilamacedo86: A comment on top of the function maybe is better in order to describe it. 

```javascript
/**
* The InMemory pubsub implementation wants an object
* etc ...
*/
```

server/lib/pubsubNotifiers/notifiers/inMemory.js
```diff
@@ -0,0 +1,25 @@
...
+    // Whereas the postgres one would expect a string
+    try {
+      payload = JSON.parse(payload)
```
camilamacedo86: It is missing `;`. Is not this project with a lint lib? Should this rule not be added there?

server/lib/resolvers/resolverMapper.js
```diff
@@ -33,14 +33,32 @@ module.exports = function (dataSources, resolverMappings) {
     let builder = resolverBuilders[dataSource.type]
 
     if (builder) {
```
camilamacedo86: Here shows that this method is missing encapsulations. Note that the complexity to understand it is high and it is doing many actions.

server/lib/resolvers/resolverMapper.js
```diff
@@ -33,14 +33,32 @@ module.exports = function (dataSources, resolverMappings) {
...
-      const {compiledRequestMapping, compiledResponseMapping} = compileMappings(resolverMapping.requestMapping, resolverMapping.responseMapping)
+      const compiledRequestMapping = compile(resolverMapping.requestMapping)
+      const compiledResponseMapping = compile(resolverMapping.responseMapping)
       if (builder.buildResolver && typeof builder.buildResolver === 'function') {
```
camilamacedo86: Should it no be better if extracted as a function?

Replace :
```javascript
if (builder.buildResolver && typeof builder.buildResolver === 'function') {
```

For
```javascript
 if ( isBuildResolverFuction(builder) ) {
```

server/lib/util/filterEvaluator.js
```diff
@@ -0,0 +1,91 @@
...
+      let operator = keys[0]
+      let args = expression[operator]
+
+      let argsType = typeof args
+
+      if (argsType === 'string' || argsType === 'number' || argsType === 'boolean' || argsType === 'undefined') {
+        return evalString(operator, context) === evalString(args, context)
+      }
+
+      if (basicOperators[operator]) {
```
camilamacedo86: Is not possible the operator has an attribute to tell this type? Then we could have here something as : 

```javascript
switch(operator.type) {
    case "loop":
        return ....
    case "basic":
        return ....
    default:
        throw Error('unkown operator ' + operator + ' in expression ' + JSON.stringify(expression))
}

```

server/lib/util/filterEvaluator.js
```diff
@@ -0,0 +1,91 @@
...
+    const arg2 = evalString(args[1], context)
+    return expressionFn(arg1, arg2)
+  }
+
+  function evalString (str, context) {
+    if (typeof str !== 'string' || str.length <= 1) {
+      return str
+    }
+    const c1 = str[0]
+    if (c1 === '$') {
```
camilamacedo86: What means `c1`? Could it have a better name to clarify its propose?

server/server.js
```diff
@@ -208,3 +210,16 @@ async function disconnectDataSources (dataSources) {
     }
   }
 }
+
+function newSubScriptionServer (server, schema) {
+  // Set up the WebSocket for handling GraphQL subscriptions
+  /* eslint-disable no-new */
```
camilamacedo86: Why the comment is not on top of function...

server/server.js
```diff
@@ -219,3 +210,16 @@ async function disconnectDataSources (dataSources) {
...
+  new SubscriptionServer({
```
aliok: We need to save a reference to current subscription server, close it, and create a new one.

This PR got very big and we should better merge it ignoring this problem. 
I created a new task for that: https://issues.jboss.org/browse/AEROGEAR-7696

server/lib/pubsubNotifiers/pubsubNotifier.js
camilamacedo86: @darahayes the problem is if we don't start the code as it should be then it will not grow as it should be. This code has many ifs, and if someone will improve it will add more ifs then if we encapsulate the logic is not required comments because what it is doing will be clear as we are doing things that allow improvements in a healthy way. Is it something that makes sense?

server/lib/pubsubNotifiers/pubsubNotifier.js
camilamacedo86: I think it should be inside of the func getPublish (PubSubClass); as described in the next comment.

server/config/index.js
darahayes: Variable declarations at the top are typically used for declaring imports.

server/lib/pubsubNotifiers/notifiers/inMemory.js
darahayes: All of this code is linted using [standard](https://standardjs.com/)

server/lib/pubsubNotifiers/notifiers/inMemory.js
darahayes: This particular comment isn't about the overall function, this comment is specifically about the `JSON.parse()` call underneath it.

server/server.js
darahayes: @aliok Okay that sounds good. In fact this whole file needs to be refactored to some extent. For example the current schema reload feature means that every time a request comes into the server, we are initializing a new GraphQL engine. This adds significant overhead to every request. I think there is a better way to do it though.

server/config/index.js
camilamacedo86: @darahayes at on top typically is declared all that is global to the program in all languages. For default, it follows the following standard/convention.

```javascript
imports ...
global vars ...
functions/methods ...
```

It is something important to make easier to keep the project maintained, global vars in many places make harder to read and understand the code as could cause mistakes, for example, declare the same value again in another global var. 

Note that this standard is valid for any language. `global variables are often available by declaring a variable at the top level of the program.` See [here](https://en.wikipedia.org/wiki/Global_variable)

Also, it is a good practice to avoid global variables in JS and try to use always local variable instead of it. Following some references.

* https://www.w3schools.com/js/js_best_practices.asp
* https://www.w3.org/wiki/JavaScript_best_practices

server/lib/pubsubNotifiers/notifiers/inMemory.js
camilamacedo86: Then, maybe it will be more clear if be as `payload = JSON.parse(payload) // required becase of  `

If the explanation is because of some line it should be in front of it or at least on top of it. Also, IMHO usually when the code requires many comments is because the code is not clear enough which means that the vars can have better names, and/or the logic can be encapsulated, and/or maybe it is unnecessarily complex.

We could start to do things more readable and easier to keep maintained as improved. I know that it is a detail and sounds boring, however, the details matters and these details at the end make all difference to read a code and etc ..

darahayes:
@aliok I am so sorry how big this PR has gotten, I have added unit tests which led me to do some refactoring around how we wrap the resolver with the publish function.

One minor additional thing I've added (and put unit tests in for) is allowing the 'magic' publish options.

It is possible to define a subscription mapping like this:

```
{
    type: 'Subscription',
    field: 'memeAdded'
    // created/updated at, SchemaId
}
```

^^ if a topic is not defined, the subscription resolver will be set up to listen on `subscriptionMapping['field']`.

And in a resolver mapping we can do this:

```
{
    type: 'Mutation',
    field: 'createMeme',
    requestMapping: `...`,
    responseMapping: '...',
    publish: 'memeAdded'
    // created/updated at, SchemaId
  }
```

And under the hood, a default publisher function is built that will publish a default payload to a topic with the same name as `resolverMapping['publish']`

This is a nice concise syntax that covers the most common use case. It is still possible to apply filtering while also using this more simplified configuration.

server/config/index.js
aliok: +1 on that. @camilamacedo86 great suggestion. I will create a GH issue for that.
I don't want this PR to get any bigger.

aliok: https://github.com/aerogear/data-sync-server/issues/33

aliok:
@darahayes verified functionality locally but only the case in memeo list schema. unit tests look good as well.
Could you please review the test cases at the integration tests JIRA? https://issues.jboss.org/browse/AEROGEAR-7691

Approving this PR.

aliok:
@david-martin 
> Another thing: we need to discuss with David and the team about syncing the sync server instances. I know we talked about it and it seems necessary, but maybe it is not in the scope right now.

What do you think about this? cc @darahayes 

This is about syncing the sync server states using the same mechanisms we have. @darahayes showed some great thinking and brought this, but we're not sure if this is something in scope.

Imagine the scenario:
* User1 is connected to server1 (websocket connection subscriptions)
* User2 is connected to server2 (websocket connection subscriptions)
* User1 creates a document which should result in publishing of a subscription event
* User2 won't get the event

darahayes:
@aliok This will work when using the postgres pubsub implementation. 

* Client 1 is connected to server 1
* Client 2 is connected to server 2
* Client 1 performs an action (e.g. create a new resource)
* Server 1 publishes this to the Postgres PubSub mechanism
* Server 1 and Server 2 receive the message and notify clients as necessary.

It's basically implemented although the postgres Pubsub implementation in this PR is really a skeleton and I haven't tried it at all.

aliok:
@darahayes 
> This will work when using the postgres pubsub implementation.

You're right. I think I got confused with the in memory pubsub.

> I would suggest adding the following steps to the integration testing ticket:

added to the ticket.

> I feel that we should focus the integration tests much more on the Postgres implementation because the In Memory one is really not for production use.

We can target having Postgres pubsub for integration tests initially and can see how easy it is to create separate tests for pubsub later.

darahayes:
@aliok Great! I have rebased now so ready to merge as long as you're happy!

PR #34, opened 2018-07-26
Adds preHook and postHook fields to resolver

josemigallas:
JIRA: https://issues.jboss.org/browse/AEROGEAR-7657

Simply adds `preHook` and `postHook` String fields to the `Resolver` sequalize model.

UI: https://github.com/aerogear/data-sync-ui/pull/64


darahayes:
Thanks @josemigallas, I'm happy with this PR. However we have #32 which is very very large and it's just been rebased in order to merge to master. If you wouldn't mind waiting until that one gets merged first. Then I will approve :)

PR #36, opened 2018-07-30
Adjust inmem to example schema

wtrocki:
New updates for the schema that involve relationships.
Supports both inmemory and posgress updates.

## Notes

Now it is not possible to create meme without Profile - user needs to be created first.

### Memeolist - In memory 

To start the application with MemeoList schema and queries with an in-memory data source, run these commands:
```
docker-compose -p aerogeardatasyncserver up
npm run db:init:memeo:inmem
npm run dev:memeo
```

### Memeolist - Postgres 

To start the application with MemeoList schema and queries with an Postgres source, run these commands:
```
docker-compose -p aerogeardatasyncserver up
npm run db:init:memeo:postgres
npm run dev:memeo
``` 

Make sure that you remove previous schema by dropping containers with postgres sql.


examples/memeolist.query.graphql
```diff
@@ -2,12 +2,52 @@ query allMemes{
...
+    owner
+    likes
+    comments {
+      id
+    }
+  }
+}
+
+query profile {
+  profile(email:"wtr@redhit.com"){
```
aliok: redhit :D

wtrocki:
Ready for review!

server/lib/resolvers/builders/nedb.js
```diff
@@ -16,6 +16,7 @@ function buildNeDBResolver (dataSource, compiledRequestMapping, compiledResponse
       const parsedQuery = JSONParse(queryString)
 
       if (parsedQuery.error) {
+        console.error('Error when parsing query', queryString, parsedQuery.error)
```
aliok: Please use "log" like in other modules. There's actually a linter rule that prevents using console

aliok:
Some problems:
* `likeMeme` returns true even when the meme doesn't exist: Both inmem and Postgres
* Same thing above, but for `postComment` and only for inmem

@wtrocki  if you have the energy to fix them it would be nice.

server/lib/resolvers/builders/nedb.js
aliok: I wonder why that linter didn't complain...

examples/memeolist.tables.sql
```diff
@@ -1,6 +1,21 @@
-DROP TABLE IF EXISTS "Meme";
+CREATE TABLE Profile (
```
darahayes: Can we add back in the following here:

```
DROP TABLE IF EXISTS "Meme";
DROP TABLE IF EXISTS "Profile";
DROP TABLE IF EXISTS "Comment";
```

It just makes local dev a lot simpler

server/lib/resolvers/builders/nedb.js
```diff
@@ -48,6 +49,7 @@ function buildNeDBResolver (dataSource, compiledRequestMapping, compiledResponse
 
       function mapResponse (err, res) {
         if (err) return reject(err)
+        if (!res) return resolve()
```
camilamacedo86: @aliok should not the linter complain about this point either?

aliok: @camilamacedo86 I rebased this branch on top of master (has now linter config fixed)  *locally* and it didn't complain about that line. What's the problem there? BTW, we're using this format: https://standardjs.com

wtrocki:
> likeMeme returns true even when the meme doesn't exist: Both inmem and Postgres

I really struggled to get this implemented in the current resolver format. 
Normally I would fetch the Meme, increment number of the likes and save it. Return true or false.
In the template I cannot execute two queries at the same time or have logic like:

```
if(numerOfRowsAffected ===0) return false;
```

> Same thing above, but for postComment and only for inmem

That should be easy fix.

aliok:
ok, good to merge then.

PR #37, opened 2018-07-30
fix: Integration Tests for Subscriptions and Publishers

darahayes:
This PR introduces tests for two main features:

* Subscriptions
* The publishing mechanism used in resolvers

In order to improve testability some refactoring has been done. The main changes are to do with server startup. I have refactored `index.js` to represent the entire service as a `SyncService` class with `start` and `stop` functions. Then inside the tests there is a subclass called `RestartableSyncService` which overrides some of default start/stop logic to allow forceful restarts of the server. This will come in very handy for testing, especially with multiple server instances.

You can see an example of this in `subscriptions.integration.test.base.js` where use two servers in the test case.


integration_test/helper.js
```diff
@@ -50,6 +51,58 @@ function Helper () {
     await new Promise(resolve => setTimeout(resolve, 1000))
     // await this.syncService.restart()
   }
+
+  this.cleanMemeolistDatabase = async (t) => {
```
camilamacedo86: Could you change the `t` for some name which clarifies better which is `t`?

integration_test/helper.js
```diff
@@ -50,6 +51,58 @@ function Helper () {
...
+
+    const client = new Client({
+      user: 'postgresql',
+      password: 'postgres',
+      database: 'memeolist_db',
+      host: memeoListDbHost,
+      port: memeoListDbPort
+    })
+
+    try {
```
camilamacedo86: IHMO it should be encapsulated in a function connect since this is the action.

integration_test/helper.js
```diff
@@ -50,6 +51,58 @@ function Helper () {
...
+      await client.connect()
+      await client.query('SELECT 1')
+    } catch (err) {
+      t.log('Unable to connect memeolist database for preparing it for the integration tests')
+      throw err
+    }
+
+    try {
```
camilamacedo86: IHMO it should be encapsulated in function prepareMemoList().
Also, in the top of the func it could have a comment gerated following the convention(it is done automatically by IDEA when you enter with `/*` ) as: 

```javascript
/******
* It will delete the tables if exists and create .... bla bla bla
* The database language here is SQL. 
*params ...
/
```

darahayes:
Pinging @aliok and @wtrocki for review.

wtrocki:
@camilamacedo86 You can participate actively in development of sync server by contributing to one available tickets: https://issues.jboss.org/secure/RapidBoard.jspa?rapidView=4283&selectedIssue=AEROGEAR-7604 
Just leave comment that you plan to work on something.

We value your opinion on reviews, however we may simply ignore them as reviews are done by code owners (Ali and Dara) and not community members.

index.js
```diff
@@ -2,20 +2,61 @@ const config = require('./server/config')
 const { log } = require('./server/lib/util/logger')
 const PubSub = require('./server/lib/pubsubNotifiers/pubsubNotifier')
 
-let { pubsubConfig, postgresConfig } = config
-let { port } = config.server
+class DataSyncService {
```
wtrocki: I do not want be another "clean code" experts, but if we extract this class to separate file then we could avoid `require.main === module`. It will be just easy and nice

integration_test/datasource.integration.test.base.js
```diff
@@ -1,4 +1,4 @@
-const {test} = require('ava')
+const { test } = require('ava')
 
 /// //////////////// NOTE /////////////////////////////////////
 /// /////////// order of tests are imporant!///////////////////
```
wtrocki: Based on what I see they are not anymore :)

integration_test/util/testApolloClient.js
```diff
@@ -0,0 +1,62 @@
...
+      const { kind, operation } = getMainDefinition(query)
+      return kind === 'OperationDefinition' && operation === 'subscription'
+    },
+    wsLink,
+    httpLink
+  )
+
+  return new ApolloClient({
+    link: link,
+    cache: new InMemoryCache()
```
wtrocki: Is that required for tests?

darahayes: This is boilerplate code taken directly from the Apollo docs to set up a client that can talk via http and via websockets. It looks really weird but I guess it's because they have support for using any http lib and any websocket lib. (For compatibility across browsers and node).

I've also created a wrapper class around the Apollo client that 'Promisifies' the subscribe method in order to make testing of subscriptions a bit simpler. See in the tests how I can do something like `const result = await subscription`

package.json
```diff
@@ -38,13 +38,23 @@
...
+    "node-fetch": "^2.2.0",
     "nodemon": "^1.17.5",
     "pre-commit": "^1.2.2",
-    "standard": "^11.0.1"
+    "standard": "^11.0.1",
+    "stoppable": "^1.0.6",
+    "ws": "^6.0.0"
   },
   "dependencies": {
+    "apollo-client": "^2.3.7",
```
aliok: It is good that you used apollo-client. But, on the other hand, we have 2 clients now apollo-fetch and apollo-client.
I say we do nothing for now but we should migrate tests using apollo-fetch to apollo-client in the future.

integration_test/util/restartableSyncService.js
```diff
@@ -6,7 +6,7 @@ class RestartableSyncService extends DataSyncService {
     this.app.server = stoppable(this.app.server, 0)
     await this.app.server.listen(this.port)
     this.log.info(`Server is now running on http://localhost:${this.port}`)
-    await new Promise(resolve => setTimeout(resolve, 1000))
+    await new Promise(resolve => setTimeout(resolve, 200))
```
camilamacedo86: IMHO in order to follow the best practices and to avoid magic numbers the `200` here should be a const with a name which clarifies its propose.

integration_test/util/restartableSyncService.js
```diff
@@ -15,7 +15,7 @@ class RestartableSyncService extends DataSyncService {
     this.app.server.stop()
     await this.initialize()
     await this.start()
-    await new Promise(resolve => setTimeout(resolve, 1000))
+    await new Promise(resolve => setTimeout(resolve, 200))
```
camilamacedo86: The same of above _IMHO in order to follow the best practices and to avoid magic numbers the 200 here should be a const with a name which clarifies its propose._ 

Note that it has been used in more than one place too.

package.json
darahayes: @aliok great point. I found this very easy to do so if you like, I can do it after your review (don't want to add more noise). 

Also, I'm just noticing that this should be a dev dependency :)

aliok:
> @camilamacedo86 You can participate actively in development of sync server by contributing to one available tickets: https://issues.jboss.org/secure/RapidBoard.jspa?rapidView=4283&selectedIssue=AEROGEAR-7604
Just leave comment that you plan to work on something.
> We value your opinion on reviews, however we may simply ignore them as reviews are done by code owners (Ali and Dara) and not community members.

good points @wtrocki.

@camilamacedo86 thanks a lot for your comments. We really appreciate your feedback!
Even though your points are mostly valid, we might ignore some of them. Ignored comments don't mean they're bad. They're perhaps not in the scope of the respective PRs or they're a bit hard to implement or they have the risk to create new bugs. Please bear with us on this.
Thanks again for the valuable comments! Nice to have you on the team!

package.json
aliok: > (don't want to add more noise).
Yeah, definitely.

> if you like, I can do it after your review 
even better :laughing:

integration_test/util/restartableSyncService.js
darahayes: @camilamacedo86 Thanks for this. This is actually great point. This is in order to avoid a timing bug (for some reason the thing isn't always initialized instantly).

I've been working on this PR for the last couple of days and I forgot to come back and tidy this up :)

aliok:
IMO, the general structure is very good. Tests are nice, handling lots of cases.
I think we're good to merge after the comments are addressed. They don't suggest big changes.

aliok:
@darahayes thanks for migrating the existing tests to apollo-client!

PR #41, opened 2018-08-01
Metrics endpoint + resolver metrics

psturc:
https://issues.jboss.org/browse/AEROGEAR-7650

### Checklist
- [x] metrics endpoint with basic metrics
- [x] resolver timing metrics (time period between request & response of database queries)
- [x] collecting number of queries/mutations made (possibly split to un/successful ones)
- [x] change the way that request information is passed to metrics
- [x] "inmemory" type resolver metrics
- [x] logging of dataSourceType (in logger & metrics)
- [x] server response metrics

### How to verify this
`npm install` (new dependency **prom-client** has been added)
`docker-compose -p aerogeardatasyncserver up`
`npm run db:init:memeo:postgres`
`npm run dev`

1. Go to http://localhost:8000/metrics and check that metrics data is available
1. Go to http://localhost:8000/graphiql, paste in following and run it:
```
mutation createprofile {
  createProfile(email: "jon@doh.com", displayname: "Jon D Oh", pictureurl: "https://cdn1.thr.com/sites/default/files/imagecache/scale_crop_768_433/2011/10/doh_a.jpg") {
    email
    id
    displayname
  }
}

query getAllMemes {
  allMemes {
    id
    likes
    comments {
      id
      owner
      comment
    }
  }
}
```
3. Refresh http://localhost:8000/metrics and check that you see a record like 
```
resolver_timing_ms{datasource_type="Postgres",operation_type="mutation",name="createProfile"} 5
resolver_timing_ms{datasource_type="Postgres",operation_type="query",name="allMemes"} 3

requests_resolved{datasource_type="Postgres",operation_type="query",path="Query.allMemes"} 1
requests_resolved{datasource_type="Postgres",operation_type="mutation",path="Mutation.createProfile"}
```

4. There should be also record about GET (`/metrics`) and POST (`/graphiql?<query>`) requests, like:
```
server_response_ms{request_type="POST",error="false"} 9
server_response_ms{request_type="GET",error="false"} 8
```
5. When you run other queries, those values should change

### DIFF
--- a/server/lib/resolvers/builders/postgres.js
+++ b/server/lib/resolvers/builders/postgres.js
@@ -1,10 +1,12 @@
 const JSONParse = require('json-parse-safe')
 const { auditLog } = require('../../../lib/util/logger')
+const { updateResolverMetrics } = require('../../../metrics')
 
-function buildPostgresResolver (dataSource, compiledRequestMapping, compiledResponseMapping) {
+function buildPostgresResolver (dataSource, compiledRequestMapping, compiledResponseMapping, resolverMapping) {
   return function resolve (obj, args, context, info) {
     return new Promise((resolve, reject) => {
       const dataSourceClient = dataSource.getClient()
+      const requestTimeStart = Date.now()
 
       const queryString = compiledRequestMapping({
         context: {
@@ -15,6 +17,9 @@ function buildPostgresResolver (dataSource, compiledRequestMapping, compiledResp
 
       dataSourceClient.query(queryString, [], (err, res) => {
         if (err) return reject(err)
+
+        // Update metrics with resolver's reponse time
+        updateResolverMetrics(resolverMapping, Date.now() - requestTimeStart)
 
         const responseString = compiledResponseMapping({
           context: {
--- a/server/lib/resolvers/builders/postgres.js
+++ b/server/lib/resolvers/builders/postgres.js
@@ -2,7 +2,7 @@ const JSONParse = require('json-parse-safe')
 const { auditLog } = require('../../../lib/util/logger')
 const { updateResolverMetrics } = require('../../../metrics')
 
-function buildPostgresResolver (dataSource, compiledRequestMapping, compiledResponseMapping, resolverMapping) {
+function buildPostgresResolver (dataSource, compiledRequestMapping, compiledResponseMapping) {
   return function resolve (obj, args, context, info) {
     return new Promise((resolve, reject) => {
       const dataSourceClient = dataSource.getClient()
@@ -19,7 +19,7 @@ function buildPostgresResolver (dataSource, compiledRequestMapping, compiledResp
         if (err) return reject(err)
 
         // Update metrics with resolver's reponse time
-        updateResolverMetrics(resolverMapping, Date.now() - requestTimeStart)
+        updateResolverMetrics(info, Date.now() - requestTimeStart)
 
         const responseString = compiledResponseMapping({
           context: {
--- a/server/lib/resolvers/builders/postgres.js
+++ b/server/lib/resolvers/builders/postgres.js
@@ -7,6 +7,7 @@ function buildPostgresResolver (dataSource, compiledRequestMapping, compiledResp
     return new Promise((resolve, reject) => {
       const dataSourceClient = dataSource.getClient()
       const requestTimeStart = Date.now()
+      info['dataSourceType'] = dataSource.type
 
       const queryString = compiledRequestMapping({
         context: {
--- a/server/metrics/index.js
+++ b/server/metrics/index.js
@@ -1,4 +1,5 @@
 const Prometheus = require('prom-client')
+const { buildPath } = require('../lib/util/logger')
 
 Prometheus.collectDefaultMetrics()
 
@@ -11,7 +12,7 @@ const resolverTimingMetric = new Prometheus.Gauge({
 const requestsResolvedMetric = new Prometheus.Counter({
   name: 'requests_resolved',
   help: 'Number of requests resolved by server',
-  labelNames: ['type']
+  labelNames: ['type', 'path']
 })
 
 exports.getMetrics = async (req, res) => {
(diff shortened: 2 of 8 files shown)


darahayes:
@psturc great work so far. I've actually noticed that our request logging middleware `express-pino` logs a response time (in milliseconds). This is the total amount of time it took for the request to finish. Under the hood it used `express-http`

Digging into the code, I see this [middleware function](https://github.com/pinojs/pino-http/blob/master/logger.js#L53-L66)

```
  function loggingMiddleware (req, res, next) {
    req.id = genReqId(req)
    req.log = res.log = logger.child({req: req})
    res[startTime] = res[startTime] || Date.now()
    if (!req.res) { req.res = res }

    res.on('finish', onResFinished)
    res.on('error', onResFinished)

    if (next) {
      next()
    }
  }
}
```

Notice how they do `res[startTime] = res[startTime] || Date.now()` and then they add listeners for `res.on('finish')` and `res.on('error').

and then in the `onResFinished` listener function:

```
  function onResFinished (err) {
    this.removeListener('error', onResFinished)
    this.removeListener('finish', onResFinished)

    var log = this.log
    var responseTime = Date.now() - this[startTime]

    // ... unrelated code
  }
```

Here they calculate the response time. This is a nice method we could use in our sever to collect things like response times and errors.

Using a middleware like this would be a nice approach so we could do something like `app.use(ourMetricsMiddleware)` and as long as that is the first middleware we load into the app, then we should get accurate readings.

In this nice example from RisingStack, they should how you you use the prometheus library to collect data with histograms, counters etc. https://github.com/RisingStack/example-prometheus-nodejs/blob/master/src/server.js

Ping me if you run into any issues while working on this 😄

aliok:
@darahayes If the requirement is finding times for HTTP requests, I would totally agree with your approach.

But, I think the requirement is finding times for resolvers. And, there can be multiple resolvers involved in a HTTP request.

Well, we can provide both :) Using the approach you mentioned to log request times and some other way to log resolver times.

psturc:
@darahayes thanks for suggestions! As Ali said, we can provide both metrics (http & resolver response times).

I think this metric would have definitely a value from client perspective (as proposed [here](https://github.com/aerogear/proposals/blob/1649150b097703f405c6b0864fca748c91381b81/sync/sdk.md#client-metrics)), but we may discuss it on the [mail thread I've started yesterday](https://groups.google.com/forum/#!topic/aerogear/OC122HNWNfo).

I'm going to implement third part of this PR, which is providing number of queries made, by type (should be quite easy)

server/lib/resolvers/builders/postgres.js
```diff
@@ -1,10 +1,12 @@
 const JSONParse = require('json-parse-safe')
 const { auditLog } = require('../../../lib/util/logger')
+const { updateResolverMetrics } = require('../../../metrics')
 
-function buildPostgresResolver (dataSource, compiledRequestMapping, compiledResponseMapping) {
+function buildPostgresResolver (dataSource, compiledRequestMapping, compiledResponseMapping, resolverMapping) {
```
aliok: @psturc I know I suggested this, but I didn't update you with the latest approach to get the type, field name etc.

If you have a look at https://github.com/aerogear/data-sync-server/blob/master/server/lib/resolvers/builders/postgres.js#L33 you will see an `info` var is passed to auditLog function.

And, here,
https://github.com/aerogear/data-sync-server/blob/master/server/lib/util/logger.js#L11
in the auditLog function, we're able to get all the information related to Graphql resolver.

I found it better to get information from `info` object like that. Could you adapt the code to get type and field name from there?

There is additional information as well, if you like to add more labels to Prom metric:
* info.operation.operation : query/mutation
* info.fieldname: field name (a. createMeme or b. memes)
* info.parentType.name: parent type (a. Mutation or b. Profile)
* info.path: the whole path for the resolver (a. Mutation.createMeme or b. allProfiles.0.memes)

In my examples above, (a) is a top level mutation.
(b) is a inner resolver (no top level) that is executed when `memes` field of a `Profile` is to be resolved.

let me know if you need any help with this.

psturc: Oh my, I have the audit logs in front of my eyes when working on this and I didn't realise! Thanks for the update! 
Should be no problem to rewrite this. I have the third part completed, so I will send suggested changes in the last commit.

aliok:
verifying again

aliok:
@psturc I just saw that there's no metrics for inmemory resolvers. Could you also update there? Same logic.

aliok:
@psturc @darahayes 
```
requests_resolved{type="mutation",path="Mutation.createMeme"} 1
resolver_timing_ms{type="mutation",name="createMeme"} 12
```
These look good. But I suggest we also add a label for the data source type (postgres/inmem/etc.). This would provide us the ability to filter data by data source and see things like this in Grafana:
* How many of my requests went to Postgres?
* How much time it took to resolve resolvers with Postgres vs Inmem.
* ...

This is a super simple code change anyway. Just add a label with a static value.

server/metrics/index.js
```diff
@@ -0,0 +1,37 @@
...
+})
+
+exports.getMetrics = async (req, res) => {
+  res.set('Content-Type', Prometheus.register.contentType)
+  res.end(Prometheus.register.metrics())
+}
+
+exports.updateResolverMetrics = (resolverInfo, responseTime) => {
+  const {
+    operation: {operation: resolverMappingType},
```
aliok: I couldn't understand where this variable comes from. Maybe too advanced JS for me :laughing:

psturc:
@aliok thanks, I'm doing a little cleanup and testing now. Will send PR soon.

server/metrics/index.js
darahayes: @aliok Two what's happening here is a sort of nested destructuring of properties from the resolverInfo Object.

You've probably unfamiliar with the names of the fields because it's internal stuff from the GraphQL engine passed as the `info` arg into the resolver function. You can see this up in the code where `updateResolverMetrics` is called from 😄

server/lib/resolvers/builders/postgres.js
```diff
@@ -7,6 +7,7 @@ function buildPostgresResolver (dataSource, compiledRequestMapping, compiledResp
     return new Promise((resolve, reject) => {
       const dataSourceClient = dataSource.getClient()
       const requestTimeStart = Date.now()
+      info['dataSourceType'] = dataSource.type
```
aliok: data source type will always be Postgres in postgres.js and InMemory in nedb.js

aliok: ah, but I see your point now. you just want to get it from the DataSource impl using its type property. that's also ok

server/metrics/index.js
```diff
@@ -0,0 +1,70 @@
...
+  const requestMethod = req.method
+
+  res['requestStart'] = Date.now()
+
+  res.on('finish', onResFinished)
+  res.on('error', onResFinished)
+
+  if (next) next()
+
+  function onResFinished (err) {
```
psturc: @darahayes @aliok maybe you've got some idea how to avoid nesting of function here?
The reason why I put it here was that I need to log `requestMethod` which wouldn't be a part of the closure if the function was declared outside of this one. Couldn't find an elegant solution.

darahayes: I'm 👍 with this approach

psturc:
@aliok @darahayes  thanks so much for suggestions and review!

PR #42, opened 2018-08-02
fix: Subscriptions Hot Reload

darahayes:
This PR ensures that the subscription server is closed and that a new one is instantiated when a schema change occurs.

It also adds an integration test to test this case.

It also makes a few minor changes in startup sequence:
	* add the unhandledRejection and unhandledException listeners at the very beginning. Because these were being added **after** initialisation + startup, any errors happening in that time would be swallowed
	* Same thing again for the stop signals listeners

### DIFF
--- a/integration_test/subscriptions.hot.reload.integration.test.js
+++ b/integration_test/subscriptions.hot.reload.integration.test.js
@@ -1,0 +1,118 @@
+const { test } = require('ava')
+const Helper = require('./helper')
+const TestApolloClient = require('./util/testApolloClient')
+const gql = require('graphql-tag')
+
+const context = {
+  apolloClient: null,
+  helper: null
+}
+
+test.before(async t => {
+  context.apolloClient = new TestApolloClient()
+  const helper = new Helper()
+
+  await helper.initialize()
+
+  // delete the all the config 1-time before starting the tests
+  await helper.deleteConfig()
+  await helper.feedConfig('complete.postgres.valid.memeo.js')
+  await helper.cleanMemeolistDatabase(t)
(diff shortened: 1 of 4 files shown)


integration_test/subscriptions.hot.reload.integration.test.js
```diff
@@ -0,0 +1,118 @@
...
+
+  // delete the all the config 1-time before starting the tests
+  await helper.deleteConfig()
+  await helper.feedConfig('complete.postgres.valid.memeo.js')
+  await helper.cleanMemeolistDatabase(t)
+  await helper.syncService.restart()
+  context.helper = helper
+})
+
+test.serial('Newly added subscriptions should be hot reloaded', async (t) => {
```
aliok: this is a great test case!

server/server.js
```diff
@@ -77,7 +78,8 @@ module.exports = async ({graphQLConfig, graphiqlConfig, postgresConfig, schemaLi
           await connectDataSources(newSchema.dataSources)
           schema = newSchema.schema
           dataSources = newSchema.dataSources
-          newSubScriptionServer(server, schema)
+          subscriptionServer.close()
+          subscriptionServer = newSubscriptionServer(server, schema)
```
aliok: the only thing I can tell is maybe close the existing subscription server at line 70 and create a new one in 2 places:
1. just after line 78 (current one)
2. just after line 88 (error case when new schema is shit and we still need to start the subscription server with the old schema) 

sorry for being this pessimistic about exceptions :)

darahayes: Good catch

PR #43, opened 2018-08-02
fix: Change schema to lower case

wtrocki:
## Motivation

Use lover case for memeo schema that will allow to execute queries without `"`


aliok:
I am ok with that.

PR #44, opened 2018-08-02
Add subscription example

wtrocki:
## Motivation

Provide examples in the form that can be used in Android and IOS client.
Still left examples for js client.
screen shot 2018-08-02 at 6 04 11 pm

### DIFF
--- a/examples/memeolist.query.examples.graphql
+++ b/examples/memeolist.query.examples.graphql
@@ -1,0 +1,62 @@
+## MEMES section
+query allMemes {
+  allMemes {
+    id
+    photourl
+    owner
+    likes
+    comments {
+      id
+      comment
+      owner
+    }
+  }
+}
+
+mutation createMeme {
+  createMeme(ownerid: "1", photourl: "https://i.imgur.com/W4dhL0b.jpg", owner: "Wojtek") {
+    id
+    photourl
+    owner
+    likes
+  }
+}
+
+mutation postComment {
+  postComment(memeid: "Fha32ppqLNWEfX3b", comment: "Like it", owner: "Wojtek") {
+    id
+    comment
+    owner
+  }
+}
+
+mutation likeMeme {
+  likeMeme(id: "Fha32ppqLNWEfX3b")
+}
+
+## Profile section
+mutation createProfile {
+  createProfile(email: "wtr1@redhit.com", displayname: "wojtek", pictureurl: "https://tinyurl.com/avataraerogear") {
+    id
+    email
+    displayname
+    pictureurl
+  }
+}
+
+query profile {
+  profile(email: "wtr1@redhit.com") {
+    id
+    email
+    displayname
+    pictureurl
+  }
+}
+
+subscription memeAdded {
+  memeAdded {
+    id
+    photourl
+    likes
+  }
+}
--- a/examples/memeolist.query.graphql
+++ b/examples/memeolist.query.graphql
@@ -1,51 +1,64 @@
-## MEMES section
-query allMemes{
-  allMemes{
-    id
-    photourl
-    owner
-    likes
-    comments {
+
+  query allMemes {
+    allMemes {
       id
-      comment
+      photourl
+      owner
+      likes
+      comments {
+        id
+        comment
+        owner
+      }
+    }
+  }
+
+  mutation createMeme($ownerid: ID!, $photourl: String!, $owner: String!) {
+    createMeme(ownerid: $ownerid, photourl: $photourl, owner: $owner) {
+        id
+        photourl
+        owner
+        likes
+    }
+  }
+
+
+  mutation postComment($memeid: ID!, $comment: String!, $owner: String!){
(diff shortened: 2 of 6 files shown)


examples/memeolist.query.examples.graphql
```diff
@@ -0,0 +1,62 @@
+## MEMES section
+query allMemes {
+  allMemes {
+    id
```
camilamacedo86: Maybe should not better change the name to `getAll`.

examples/memeolist.query.graphql
```diff
@@ -1,51 +1,64 @@
...
-    comments {
+
+  query allMemes {
+    allMemes {
       id
-      comment
+      photourl
       owner
+      likes
+      comments {
```
camilamacedo86: IMHO should be all in the singular.

examples/memeolist.variables.json
```diff
@@ -0,0 +1,9 @@
+{ 
+  "ownerid": "1", 
+  "photourl": "https://i.imgur.com/W4dhL0b.jpg", 
+  "owner": "mmartini", 
+  "email":"mmartin@rodhut.com", 
+  "displayname": "Martin Martinski", 
+  "pictureurl": "https://tinyurl.com/avataraerogear",
```
camilamacedo86: May, the variables could not be `pictureUrl` instead of `pictureurl` and `displayName` instead of `displayname`? What do you think?

server/config/index.js
```diff
@@ -17,6 +18,16 @@ if (process.env.GRAPHIQL_QUERY_FILE) {
   }
 }
 
+if (process.env.GRAPHIQL_VARIABLES_FILE) {
+  try {
+    graphiqlVariableFileContent = fs.readFileSync(process.env.GRAPHIQL_VARIABLES_FILE, 'utf-8')
+    graphiqlVariableFileContent = JSON.parse(graphiqlVariableFileContent)
```
camilamacedo86: The encoding type may be used in others place and could be changed, in this way, what do think about centralized it in a var ?

sequelize/seeders/memeolist-example-postgres.js
```diff
@@ -28,7 +28,7 @@ const resolvers = [
     field: 'comments',
     DataSourceId: 1,
     GraphQLSchemaId: 1,
-    requestMapping: `SELECT * FROM comment WHERE memeid='{{context.parent.id}}'`,
+    requestMapping: `SELECT * FROM comment WHERE memeid='{{context.parent.id}}' ORDER BY id DESC`,
```
camilamacedo86: Shows that it will not work because the field name is `memeId` and not `memeid`

examples/memeolist.query.examples.graphql
aliok: @camilamacedo86 The names come from here: https://github.com/aerogear/proposals/blob/master/dogfood.md#current-schema

but, they're not written on stone. @wtrocki feel free to rename everything, if you want to do it.

aliok: Ah, one sec: I was talking about the query name defined in the schema. But, you get the idea

server/config/index.js
aliok: For now, we can create a constant in the same file as I don't think we don't have any other place that we use an encoding.

examples/memeolist.variables.json
aliok: I actually had the initial schema with camel case. But field names are converted to lowercase because you need to wrap column names in quotes for Postgres case-sensitivity. @wtrocki is it a lot of work to fix the casing and making everything camel case? Any side effects on client side?

examples/memeolist.query.examples.graphql
wtrocki: @camilamacedo86 This is convention that team was using so far. This change shows as new but in fact clients are already implementing this. The most important pattern in graphql is to not change schema.

wtrocki:
> @wtrocki is it a lot of work to fix the casing and making everything camel case? Any side effects on client side?

Reason why I used lowercase names is that for the moment I want to avoid mapping this data from dataset to the graphql. Current query is just `SELECT * FROM X` if we want to go with different field names then I will need to do `SELECT displayname as displayName from X` etc. This is pretty much work in progress to enable us to experiment. It's definitely not the target example schema. What I'm trying to enable is quick reference  for potential users to write this to avoid boilerplate. Postgress works differently depending if system is case sensitive so the best way will be to use lower case for now especially that this is just example schema that is not part of the delivery.

aliok:
@wtrocki `SELECT "displayName" from X` would also work. Postgres column names and table names becomes case sensitive in queries when they're wrapped in double quotes.

aliok:
@wtrocki 
> @wtrocki SELECT "displayName" from X would also work. Postgres column names and table names becomes case sensitive in queries when they're wrapped in double quotes.

Well IMO, we can merge with lowercase field/column names for now.

PR #45, opened 2018-08-03
task: Add standard github files

camilamacedo86:
# WHAT
Add the standard files for github

# WHY 
It will be helpful for contributors be able to follow the guidelines

# HOW:
- Adding files with templates to open PR's and Issues as the guidelines

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,76 @@
+# How to contribute
+
+Thank you for your interest in contributing to the Aerogear project. We want
+keep this process as easy as possible so we've outlined a few guidelines below. 
+For more information about the Aerogear project, visit 
+[our website](http://aerogear.org/).
+
+## Asking for help
+
+Whether you're contributing a new feature or bug fix, or simply submitting a
+ticket, the Aerogear team is available for technical advice or feedback. 
+You can reach us at #aerogear on [Freenode IRC](https://freenode.net/) or the 
+[aerogear-dev list](http://lists.jboss.org/pipermail/aerogear-dev/)
+-- both are actively monitored.
+
+## Getting started
+
+* Make sure you have a [JIRA account](https://issues.jboss.org)
+* Make sure you have a [GitHub account](https://github.com/signup/free)
+* Submit a ticket for your issue to the 
+[Aerogear project](https://issues.jboss.org/projects/FH/), assuming one does 
+not already exist.
+  * Clearly describe the issue including steps to reproduce when it is a bug.
+  * Make sure you fill in the earliest version that you know has the issue.
+* Fork the repository on GitHub.
+
+## Making changes
+
+* Create a topic branch from where you want to base your work.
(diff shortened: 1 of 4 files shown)


.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,76 @@
+# How to contribute
+
+Thank you for your interest in contributing to the Aerogear project. We want
```
aliok: Could you please fix the casing for Aerogear. It should be AeroGear

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,76 @@
...
+keep this process as easy as possible so we've outlined a few guidelines below. 
+For more information about the Aerogear project, visit 
+[our website](http://aerogear.org/).
+
+## Asking for help
+
+Whether you're contributing a new feature or bug fix, or simply submitting a
+ticket, the Aerogear team is available for technical advice or feedback. 
+You can reach us at #aerogear on [Freenode IRC](https://freenode.net/) or the 
+[aerogear-dev list](http://lists.jboss.org/pipermail/aerogear-dev/)
```
aliok: we should get rid of aerogear-dev list as we moved away from it. Please mention the Google group though

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,76 @@
...
+-- both are actively monitored.
+
+## Getting started
+
+* Make sure you have a [JIRA account](https://issues.jboss.org)
+* Make sure you have a [GitHub account](https://github.com/signup/free)
+* Submit a ticket for your issue to the 
+[Aerogear project](https://issues.jboss.org/projects/FH/), assuming one does
```
aliok: https://issues.jboss.org/projects/FH/ is not used anymore.

aliok:
@camilamacedo86 
I added a few comments and I can add a few more but I don't know if it makes sense.
Because I think you just took these files from other projects and these issues should be fixed in a centralized place and should then be fixed in every repo we have.

I am ok to merge this and then start a discussion around this in the ML.

PS: ignore the integration test: CircleCI seems to fail sometimes when installing NPM deps.

camilamacedo86:
@aliok really thank you for the reviews .. My idea is to update/sent a PR for each main repo for we have an standard than please feel free to check it further.

camilamacedo86:
Ignore for now.

aliok:
@camilamacedo86 +1 on that approach. thanks!

aliok:
@camilamacedo86 well, we can still merge this to have PR templates

camilamacedo86:
@aliok you are the boss !!! .. If you would like to merge it now without this changes and after all discussion and centralization of the info we update this repo please feel free.

aliok:
@camilamacedo86  yeah let's do it that way.

camilamacedo86:
@aliok I create a repo with all and some changes. See here https://github.com/camilamacedo86/standards. However, I do not have permission to create a repo in the AeroGear org.

PR #46, opened 2018-08-03
Add pre and post hook

danielpassos:
Related issues:

* [AEROGEAR-7658](https://issues.jboss.org/browse/AEROGEAR-7658)

### DIFF
--- a/server/lib/resolvers/resolverMapper.js
+++ b/server/lib/resolvers/resolverMapper.js
@@ -2,6 +2,7 @@ const _ = require('lodash')
 const resolverBuilders = require('./builders')
 const { compile } = require('./compiler')
 const { wrapResolverWithPublish } = require('./wrapResolverWithPublisher')
+const { wrapResolverWithHooks } = require('./wrapResolverWithHooks')
 
 module.exports = function (dataSources, resolverMappings, pubsub) {
   const resolvers = {}
@@ -49,6 +50,8 @@ module.exports = function (dataSources, resolverMappings, pubsub) {
       compiledRequestMapping,
       compiledResponseMapping
     )
+
+    resolver = wrapResolverWithHooks(resolver, resolverMapping)
 
     // If a publish option is specified we wrap the resolver function
     if (resolverMapping.publish && !_.isEmpty(resolverMapping.publish)) {
--- a/server/lib/resolvers/resolverMapper.js
+++ b/server/lib/resolvers/resolverMapper.js
@@ -2,6 +2,7 @@ const _ = require('lodash')
 const resolverBuilders = require('./builders')
 const { compile } = require('./compiler')
 const { wrapResolverWithPublish } = require('./wrapResolverWithPublisher')
+const axios = require('axios')
 const { wrapResolverWithHooks } = require('./wrapResolverWithHooks')
 
 module.exports = function (dataSources, resolverMappings, pubsub) {
@@ -52,7 +53,7 @@ module.exports = function (dataSources, resolverMappings, pubsub) {
     )
 
     if (resolverMapping.preHook || resolverMapping.postHook) {
-      resolver = wrapResolverWithHooks(resolver, resolverMapping)
+      resolver = wrapResolverWithHooks(resolver, resolverMapping, axios)
     }
 
     // If a publish option is specified we wrap the resolver function
--- a/server/lib/resolvers/resolverMapper.js
+++ b/server/lib/resolvers/resolverMapper.js
@@ -51,7 +51,9 @@ module.exports = function (dataSources, resolverMappings, pubsub) {
       compiledResponseMapping
     )
 
-    resolver = wrapResolverWithHooks(resolver, resolverMapping)
+    if (resolverMapping.preHook || resolverMapping.postHook) {
+      resolver = wrapResolverWithHooks(resolver, resolverMapping)
+    }
 
     // If a publish option is specified we wrap the resolver function
     if (resolverMapping.publish && !_.isEmpty(resolverMapping.publish)) {
--- a/server/lib/resolvers/wrapResolverWithHooks.js
+++ b/server/lib/resolvers/wrapResolverWithHooks.js
@@ -1,15 +1,15 @@
 const _ = require('lodash')
 const {log} = require('../util/logger')
-const request = require('request')
+const axios = require('axios')
 
 exports.wrapResolverWithHooks = function wrapResolverWithHooks (resolverFn, resolverMapping) {
   return (obj, args, context, info) => {
     return new Promise(async (resolve, reject) => {
       if (resolverMapping.preHook && !_.isEmpty(resolverMapping.preHook)) {
-        log.info(resolverMapping.preHook)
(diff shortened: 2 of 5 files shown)


danielpassos:
@aliok  @darahayes @wtrocki @pb82 It is not done yet because I need to add some test but I really appreciate a first eye to know if I'm going in the correct way

sequelize/seeders/memeolist-example-postgres.js
```diff
@@ -40,6 +40,8 @@ const resolvers = [
     GraphQLSchemaId: 1,
     requestMapping: 'SELECT * FROM meme ORDER BY id DESC',
     responseMapping: '{{ toJSON context.result }}',
+    preHook: 'http://www.google.com',
```
wtrocki: I will setup this on mutation instead of query.
Additionally we can use environment variable here in order to override that in example. see that request is made

server/lib/resolvers/wrapResolverWithHooks.js
```diff
@@ -0,0 +1,36 @@
...
+  return wrapResolverWithMappingHooks(resolver, resolverMapping)
+}
+
+function wrapResolverWithMappingHooks (resolverFn, resolverMapping) {
+  return (obj, args, context, info) => {
+    return new Promise(async (resolve, reject) => {
+      if (resolverMapping.preHook && !_.isEmpty(resolverMapping.preHook)) {
+        log.info(resolverMapping.preHook)
+        try {
+          request.get(resolverMapping.preHook)
```
wtrocki: Checked that and try catch is not going to work here. We should use 
```.on('error', function(err) {
    log.err(err)
  })
```

server/lib/resolvers/wrapResolverWithHooks.js
```diff
@@ -0,0 +1,36 @@
...
+        const result = await resolverFn(obj, args, context, info)
+        resolve(result)
+      } catch (error) {
+        log.error(error)
+        reject(error)
+      }
+    }).then(function (result) {
+      if (resolverMapping.postHook && !_.isEmpty(resolverMapping.postHook)) {
+        log.info(resolverMapping.postHook)
+        request.get(resolverMapping.postHook)
```
wtrocki: add `on` as above.

wtrocki:
@danielpassos awesome change. Minor comments added. It works!

server/lib/resolvers/wrapResolverWithHooks.js
```diff
@@ -0,0 +1,36 @@
+const _ = require('lodash')
+const {log} = require('../util/logger')
+const request = require('request')
+
+exports.wrapResolverWithHooks = function wrapResolverWithHooks (resolver, resolverMapping) {
+  return wrapResolverWithMappingHooks(resolver, resolverMapping)
+}
+
+function wrapResolverWithMappingHooks (resolverFn, resolverMapping) {
```
pb82: The function has the same signature as `wrapResolverWithHooks`, why not just put everything in there?

server/lib/resolvers/wrapResolverWithHooks.js
pb82: Do we want to send any payload to the webhook?

aliok: We actually have very few callbacks in the code base and await/async with try/catch is the preferred way, but we don't want to block execution here. So, +1 on @wtrocki 's suggestion.

aliok:
IMO the direction where this PR goes is good.

server/lib/resolvers/resolverMapper.js
```diff
@@ -50,6 +51,8 @@ module.exports = function (dataSources, resolverMappings, pubsub) {
       compiledResponseMapping
     )
 
+    resolver = wrapResolverWithHooks(resolver, resolverMapping)
```
darahayes: IMO this should be something like

```js
if (resolverMapping.preHook || resolverMapping.postHook) {
  resolver = wrapResolverWithHooks(resolver, resolverMapping)
}
```

Otherwise we are needlessly adding an additional layer of function calls and try catch blocks for every request. I know it's a minor thing but it can be easily avoided.

You can still perform the `isEmpty` checks within the `wrapResolverWithHooks` function.

server/lib/resolvers/wrapResolverWithHooks.js
darahayes: I know it seems like a minor thing right now but I would suggest using a Promise based http lib such as [Axios](https://github.com/axios/axios) instead of `request`.

The reason is because request uses older conventions that are not considered good practice any more (callbacks with 3 args), which makes it impossible to use with features like async/await.

Right now it makes no difference to the code so you might ask why would I care that much? If we ever need to add any other http related functionality later (such as HTTP resolvers) the person who implements that will be inclined to use whichever library gets added as part of this PR. request will cause problems with code maintainability in the future.

server/lib/resolvers/wrapResolverWithHooks.js
darahayes: 👍

server/lib/resolvers/wrapResolverWithHooks.js
aliok: that's a good point. Let's use a promise based lib even thought we don't need it right now.

sequelize/seeders/memeolist-example-postgres.js
danielpassos: This is only for test and going to be removed

server/lib/resolvers/wrapResolverWithHooks.js
danielpassos: Because I had other things when I have created it. Going to move since it doesn't make sense anymore!

server/lib/resolvers/wrapResolverWithHooks.js
wtrocki: @danielpassos request should use post method and for preHook it should send variables and for postHook should send response as it's body.

server/lib/resolvers/wrapResolverWithHooks.js
```diff
@@ -0,0 +1,32 @@
+const _ = require('lodash')
+const {log} = require('../util/logger')
+const axios = require('axios')
+
+exports.wrapResolverWithHooks = function wrapResolverWithHooks (resolverFn, resolverMapping) {
+  return (obj, args, context, info) => {
+    return new Promise(async (resolve, reject) => {
+      if (resolverMapping.preHook && !_.isEmpty(resolverMapping.preHook)) {
+        axios.get(resolverMapping.preHook)
```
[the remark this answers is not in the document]
darahayes: for prehook we should do this:

```
axios.post(resolverMapping.preHook, { args })
```

and for posthook:

```
axios.post(resolverMapping.postHook, { result })
```

@danielpassos can you make those small changes I described and then we get this merged?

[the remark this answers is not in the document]
danielpassos: Done!

PR #47, opened 2018-08-07
fix: graphiql subscriptions

darahayes:
This PR fixes the following:

* `subscriptionsClient.subscribe() is not a function` error seen in graphiql. It turns out that a minor release of `apollo-server-express` (from 1.3.6 to 1.4.0) [broke graphiql](https://github.com/graphql-dotnet/server/issues/125#issuecomment-405704586). The fix was to downgrade back to 1.3.6 and to also explicitly add `apollo-server-module-graphiql@1.3.4` as a dependency.
* memeolist sql script was broken (`npm run db:init:memeo:postgres` failed)
* example query file was using the wrong casing (`photoUrl` vs. the correct `photourl`)


wtrocki:
Verifying now

wtrocki:
@darahayes Subscription is not included in example. Do you think it's good to have it there as well?

wtrocki:
Init script
```
== memeolist-example-postgres: migrating =======
== memeolist-example-postgres: migrated (0.021s)

DROP TABLE
DROP TABLE
DROP TABLE
CREATE TABLE
CREATE TABLE
```
Subscription
```
subscription memeAdded {
  memeAdded {
    photourl
  }
}
```
Result
```
[object Object]
```

PR #48, opened 2018-08-08
feat: add code coverage reporting

darahayes:
This PR adds an additional step to the CI config which runs both integration and unit tests and outputs code coverage to Coveralls.io

You can see the project in coveralls here: https://coveralls.io/github/aerogear/data-sync-server


wtrocki:
verification:
- Checked coveralls account 👍 
- Checked if reports files are picked up and include all the tests 👍 
- Overall impression 💯

PR #54, opened 2018-08-15
Top level auth config for server

wtrocki:
## Motivation

Branch that is going to be used for collaboration for 

NOTES: Pushed helper files that should not land into master only to help and point out preconfigured application cluster. They going to be removed after

## Verification steps

Verification will focus on memeolist example.
We will need to run memeolist without keycloak 

1. `npm run dev:memeo` >  
2. Log should appear: Keycloak authentication is not configured
3. UI queries should work

Memeolist with keycloak.
1. `npm run dev:keymemeo` >  Initializing Keycloak authentication
2. UI queries should *not* work - keycloak login page should appear (if you provide redirect url :)


aliok:
The direction of this PR looks good.

wtrocki:
@darahayes Actually you may be right. We may not need session for our needs. It's up to investigation.
Going to use this with simple end to end server to give us prototype that we can extend over the time. 
This code is nowhere close to be merged, but it is going to be used to facilitate work on clients and UI

wtrocki:
Verified that this is working and waiting for @StephenCoady merge to integrate this into Apollo 2.0

server/config/index.js
```diff
@@ -28,17 +28,19 @@ if (process.env.PLAYGROUND_VARIABLES_FILE) {
   }
 }
 
+const topLevelGraphqlPath = '/graphql'
 const port = process.env.HTTP_PORT || '8000'
 
 const config = {
   server: {
+    apiPath: topLevelGraphqlPath,
```
darahayes: Can you rename this to `graphqlEndpoint` and move it under the graphqlConfig object?

server/config/index.js
```diff
@@ -28,17 +28,19 @@ if (process.env.PLAYGROUND_VARIABLES_FILE) {
   }
 }
 
+const topLevelGraphqlPath = '/graphql'
```
darahayes: Can you rename this to `graphqlEndpoint`?

PR #57, opened 2018-08-17
updated apollo version, added config for playground instead of graphiql

StephenCoady:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-7787

## What
Updated the Apollo version
Changed the config so that playground could use the old graphiql config

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "release:validate": "./scripts/validateRelease.sh",
     "db:init": "FORCE_DROP=true node ./scripts/sync_models",
     "db:init:memeo:inmem": "FORCE_DROP=true node ./scripts/sync_models && sequelize db:seed --seed memeolist-example-inmem.js",
-    "db:init:memeo:postgres": "FORCE_DROP=true node ./scripts/sync_models && sequelize db:seed --seed memeolist-example-postgres.js && docker exec aerogeardatasyncserver_postgres_memeo_1 psql -U postgres -d memeolist_db -f /tmp/examples/memeolist.tables.sql",
+    "db:init:memeo:postgres": "FORCE_DROP=true node ./scripts/sync_models && sequelize db:seed --seed memeolist-example-postgres.js && docker exec datasyncserver_postgres_memeo_1 psql -U postgres -d memeolist_db -f /tmp/examples/memeolist.tables.sql",
     "db:shell": "docker exec -it aerogeardatasyncserver_postgres_1 psql -U postgresql -d aerogear_data_sync_db",
     "db:shell:memeo": "docker exec -it aerogeardatasyncserver_postgres_memeo_1 psql -U postgresql -d memeolist_db"
   },
@@ -63,7 +63,7 @@
     "ws": "^6.0.0"
   },
   "dependencies": {
-    "apollo-server-express": "1.3.6",
+    "apollo-server-express": "^2.0.2",
     "apollo-server-module-graphiql": "1.3.4",
     "axios": "^0.18.0",
     "body-parser": "^1.18.3",
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "test:cover": "nyc --reporter=text-lcov ava --concurrency=1 '*.test.js' '**/*.test.js' | coveralls",
     "start": "node ./index.js",
     "dev": "nodemon ./index.js | pino-pretty --colorize --translateTime",
-    "dev:memeo": "GRAPHIQL_QUERY_FILE='examples/memeolist.query.graphql' GRAPHIQL_VARIABLES_FILE='examples/memeolist.variables.json' nodemon ./index.js | pino-pretty --colorize --translateTime",
+    "dev:memeo": "PLAYGROUND_QUERY_FILE='examples/memeolist.query.graphql' PLAYGROUND_VARIABLES_FILE='examples/memeolist.variables.json' nodemon ./index.js | pino-pretty --colorize --translateTime",
     "lint": "eslint .",
     "format": "eslint . --fix",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
     "release:validate": "./scripts/validateRelease.sh",
     "db:init": "FORCE_DROP=true node ./scripts/sync_models",
     "db:init:memeo:inmem": "FORCE_DROP=true node ./scripts/sync_models && sequelize db:seed --seed memeolist-example-inmem.js",
-    "db:init:memeo:postgres": "FORCE_DROP=true node ./scripts/sync_models && sequelize db:seed --seed memeolist-example-postgres.js && docker exec aerogeardatasyncserver_postgres_memeo_1 psql -U postgres -d memeolist_db -f /tmp/examples/memeolist.tables.sql",
+    "db:init:memeo:postgres": "FORCE_DROP=true node ./scripts/sync_models && sequelize db:seed --seed memeolist-example-postgres.js && docker exec datasyncserver_postgres_memeo_1 psql -U postgres -d memeolist_db -f /tmp/examples/memeolist.tables.sql",
```
aliok: If we're changing the pod name here, we also need to update the docker compose command (currently `docker-compose -p aerogeardatasyncserver up -d`). I think docker compose command is only mentioned in README.

By default, docker compose prefixes the pod names with the directory name unless `-p` param is used. I think you had a different directory name and you didn't use `-p` and you had trouble.

server/server.js
```diff
@@ -1,43 +1,58 @@
...
-    return graphql(req, res, next)
+  let apolloServer = new ApolloServer({
+    schema,
+    context: async ({ req }) => {
+      return {
+        request: req
+      }
+    },
+    tracing,
+    playground: {
```
aliok: Please update the Graphiql related part in README.

Playground URL is http://localhost:8000/graphql

aliok:
I got this message when I tried hot reload:

```
{"level":50,"time":1534496335408,"msg":"[object Promise]","pid":6874,"hostname":"dhcppc14","type":"Error","stack":"TypeError: Cannot read property 'endpoint' of undefined\n    at newExpressApp (/home/aliok/Desktop/code-redhat/aerogear/aerogear-data-sync-server/server/server.js:37:38)\n    at onReceive (/home/aliok/Desktop/code-redhat/aerogear/aerogear-data-sync-server/server/server.js:90:24)\n    at <anonymous>","v":1}
```

Here is how I trigger hot reload:
Create a new file in the root of the source tree and run it directly:

```
process.env.POSTGRES_USERNAME = 'postgresql'
process.env.POSTGRES_PASSWORD = 'postgres'
process.env.POSTGRES_DATABASE = 'aerogear_data_sync_db'
process.env.POSTGRES_HOST = '127.0.0.1'
process.env.POSTGRES_PORT = '5432'

const config = require('./server/config')
let {postgresConfig} = config
console.log(postgresConfig)

const PGPubsub = require('pg-pubsub')

var pubsubInstance = new PGPubsub({
  user: postgresConfig.user,
  host: postgresConfig.host,
  database: postgresConfig.database,
  password: postgresConfig.password,
  port: postgresConfig.port
})

pubsubInstance.publish('aerogear-data-sync-config', {})
```

aliok:
Good work @StephenCoady  . Thanks for taking care of this!

Left some small comments.

package.json
aliok: 2nd thought: let's revert this change :)

server/server.js
```diff
@@ -72,8 +87,10 @@ module.exports = async ({graphQLConfig, graphiqlConfig, postgresConfig, schemaLi
 
       if (newSchema) {
         schema = newSchema.schema
-        subscriptionServer.close()
-        subscriptionServer = newSubscriptionServer(server, schema)
+        const newApp = newExpressApp(schema, server)
```
darahayes: I think integration tests are failing because the tracing and playground config aren't being passed here is all.

StephenCoady: Yep just spotted the same. Fixed now and testing again. Making some more changes will ping you shortly. Thanks.

server/server.js
```diff
@@ -1,43 +1,58 @@
...
 const schemaParser = require('./lib/schemaParser')
 const schemaListenerCreator = require('./lib/schemaListeners/schemaListenerCreator')
 
-module.exports = async ({graphQLConfig, graphiqlConfig, postgresConfig, schemaListenerConfig}, models, pubsub) => {
-  const {tracing} = graphQLConfig
-  let {schema, dataSources} = await buildSchema(models, pubsub)
-  await connectDataSources(dataSources)
-
-  const app = express()
+function newExpressApp (schema, httpServer, tracing, playgroundConfig) {
```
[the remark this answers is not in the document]
StephenCoady: what about just init?

server/config/index.js
```diff
@@ -37,11 +37,11 @@ const config = {
   graphQLConfig: {
     tracing: true
   },
-  graphiqlConfig: {
-    endpointURL: '/graphql', // if you want GraphiQL enabled
+  playgroundConfig: {
+    endpoint: '/graphql', // if you want GraphiQL enabled
```
wtrocki: This is for consumer UI to call existing server. 
It looks like with new version users have no decision over the name of graphql server endpoint. 
Am I right?

darahayes: This is actually the config that specifies where the playground should be served. The default is `/graphql` though. We found some issues when we tried other endpoints though.

darahayes:
@StephenCoady don't worry about this code coverage stuff. I think I'm going to switch to a different code coverage provider.

PR #58, opened 2018-08-17
Fix graphiql URL

danielpassos:


darahayes:
Hey @danielpassos since we migrated to Apollo v2, we do not use GraphiQL anymore. We use something called [GraphQL Playground](https://www.apollographql.com/docs/apollo-server/features/graphql-playground.html). GraphQL Playground is served from the `/graphql` endpoint.

PR #59, opened 2018-08-20
Remove docker-compose -p parameter on README

danielpassos:


darahayes:
Can we maybe remove the `-d` options? They are pretty opinionated. This should be something a more advanced user decides for themselves.

PR #62, opened 2018-08-21
Add voter user and role to sample memeo realm

wtrocki:
## Motivation
 
Add realm user and realm role to sample keycloak file. 
Updates were made also on security server.


wtrocki:
WOW. I merged this change as I got false impression that it was approved. Please let me know if you have comments on this (we can revert/improve stuff on master as it's not affecting the code any way)

darahayes:
@wtrocki It's all good just double check in the future 😉

PR #68, opened 2018-08-23
Add login middleware

wtrocki:
## Motivation

Allow to use playground with keycloak

## Current flow

1. localhost:8000/login
2. Login to keycloak
3. Copy token
4. Use token in the playground

http://www.giphy.com/gifs/ukgZfagWxYxkhqIQh2

## Investigation

Refresh playground tab with token after login (it's attached to req.session)

### DIFF
--- a/server/security/index.js
+++ b/server/security/index.js
@@ -25,14 +25,19 @@ exports.applyAuthMiddleware = (keycloakConfig, expressRouter, apiPath) => {
     }, keycloakConfig)
 
     // Install general keycloak middleware
-    expressRouter.use(keycloak.middleware())
+    expressRouter.use(keycloak.middleware({
+      admin: apiPath
+    }))
 
     // Protect the main route for all graphql services
     // Disable unauthenticated access
     expressRouter.use(apiPath, keycloak.protect())
 
     expressRouter.get('/login', keycloak.protect(), function (req, res) {
-      const token = JSON.stringify(JSON.parse(req.session['keycloak-token']))
+      let token = req.session['keycloak-token']
+      if (token) {
+        token = JSON.parse(token).access_token
+      }
       res.send({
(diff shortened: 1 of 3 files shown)


wtrocki:
Ready for review!

server/security/index.js
```diff
@@ -30,6 +30,13 @@ exports.applyAuthMiddleware = (keycloakConfig, expressRouter, apiPath) => {
     // Protect the main route for all graphql services
     // Disable unauthenticated access
     expressRouter.use(apiPath, keycloak.protect())
+
+    expressRouter.get('/login', keycloak.protect(), function (req, res) {
+      const token = JSON.stringify(JSON.parse(req.session['keycloak-token']))
+      res.send({
```
darahayes: If you change this to the following, it's a little bit handier.

```js
const token = JSON.parse(req.session['keycloak-token'])
res.json({ authorization: `Bearer ${token.access_token}`})
```

wtrocki: Yeah. done that already. Just haven't pushed :D

server/security/index.js
```diff
@@ -25,11 +25,23 @@ exports.applyAuthMiddleware = (keycloakConfig, expressRouter, apiPath) => {
...
     // Disable unauthenticated access
     expressRouter.use(apiPath, keycloak.protect())
+
+    expressRouter.get('/login', keycloak.protect(), function (req, res) {
+      let token = req.session['keycloak-token']
+      if (token) {
+        token = JSON.parse(token).access_token
+      }
+      res.send({
+        'Authorization': 'Bearer ' + token
```
darahayes: Shouldn't this be token.access_token?

darahayes: Also `res.json` is better here

PR #74, opened 2018-08-28
added local development flow for keycloak

StephenCoady:
## Motivation
Added a small readme section for using Keycloak for local development

https://issues.jboss.org/browse/AEROGEAR-7839

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,24 +135,22 @@ for [Memeolist](#whats-memeolist) sample application.
 
 ## Using Keycloak for local development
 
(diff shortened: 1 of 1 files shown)


darahayes:
@StephenCoady great work! I've just realised:

* This is specifically for memeolist
* There is a small section on Keycloak integration with memeolist under the memeolist heading (further down in the Readme)

Your PR adds extra info which is really handy though. Would you be able to merge your section with the section found under the memeolist section? Thanks!

StephenCoady:
@darahayes I split out the info in my original commit so that the relevant info is under each heading. let me know if its ok.

PR #77, opened 2018-08-28
Make owner a relationship on Meme

danielpassos:


examples/memeolist.tables.sql
```diff
@@ -12,9 +12,8 @@ CREATE TABLE profile (
 CREATE TABLE meme (
   id SERIAL NOT NULL PRIMARY KEY,
   photourl CHARACTER VARYING(500) NOT NULL,
-  owner CHARACTER VARYING(100) NOT NULL,
   likes NUMERIC NOT NULL,
-  ownerid SERIAL NOT NULL references profile(id)
```
darahayes: Really happy with this PR except this one thing. I strongly believe this field should still be `ownerid` or `ownerId`. (whereas at a higher level, the resolved object will be called `owner`)

examples/memeolist.query.examples.graphql
```diff
@@ -1,10 +1,14 @@
-## MEMES section
 query allMemes {
```
wtrocki: This file is no longer used. We could remove it.

examples/memeolist.tables.sql
wtrocki: Resolvers are now overly simplified (`select * from meme`) so table and graphql names are connected.

wtrocki:
Great improvement!

examples/memeolist.tables.sql
darahayes: Okay I will concede that one 😄

examples/memeolist.query.examples.graphql
darahayes: Can we keep it just for now? I've had some issues with the playground around variables. It tends to be a little buggy sometimes so I'd like to hold onto this file just for a short time

wtrocki: Yeah! That was the reason to keep it.
We need variable format as server is now source of truth and we get those queries back to mobile,
but having separate one with variables is also good, but brings more maintenance cost.

wtrocki:
@ziccardi To use it see: https://github.com/aerogear/data-sync-server#postgres-1

PR #78, opened 2018-08-29
added security and validation to apollo server

StephenCoady:
added sane defaults and the ability to configure the values.

## Motivation
https://issues.jboss.org/browse/AEROGEAR-7721
https://issues.jboss.org/browse/AEROGEAR-7719

## What
Added some security features to the sync server to validate queries based on depth, complexity and also added a configurable option to allow or disallow introspection.

## Why
Depth limiting lowers the risk of cyclical requests which could be used as a DOS attack. Complexity ensures the server runs at a healthy level. The ability to disable Introspection means increased security.

## How
Using: 
https://github.com/4Catalyzer/graphql-validation-complexity
https://github.com/stems/graphql-depth-limit

## Verification Steps
1. Run the server from this branch. 
1. In the playground create multiple queries and run them so that there are no errors.
1. In the .env file, set INTROSPECTION to false.
1. In the playground you should now see errors as playground uses introspection to gather schema information.
1. Reset the INTROSPECTION env var.
1. Set the DEPTH_LIMIT env var to 1. This should mean all queries are denied as any returned data has a depth of at least 1.
1. Increase the depth gradually and verify that the results of the queries match what is expected.
1. Repeat the above 3 steps for complexity, setting it to a low value to verify it works.

## Progress
- [x] Add validation to server
- [x] Add introspection
- [x] Add sane defaults
- [x] Allow values to be easily configurable
- [ ] Add cost directives


server/config/index.js
```diff
@@ -57,7 +57,12 @@ const config = {
     type: null, // e.g. type 'keycloak' or 'passport'
     config: null // implementation specific config
   },
-  schemaListenerConfig: undefined
+  schemaListenerConfig: undefined,
+  serverSecurity: {
+    introspection: process.env.INTROSPECTION || true,
+    depthLimit: process.env.DEPTH_LIMIT || 10,
```
wtrocki: Tested that with DEPTH_LIMIT=1

darahayes: I would maybe rename this to `QUERY_DEPTH_LIMIT`

server/config/index.js
```diff
@@ -57,7 +57,12 @@ const config = {
     type: null, // e.g. type 'keycloak' or 'passport'
     config: null // implementation specific config
   },
-  schemaListenerConfig: undefined
+  schemaListenerConfig: undefined,
+  serverSecurity: {
+    introspection: process.env.INTROSPECTION || true,
```
darahayes: I actually think we can totally remove the introspection stuff altogether. One thing you will notice is if you set `NODE_ENV=production` then introspection and the playground are automatically disabled.

server/apolloServer.js
```diff
@@ -9,6 +11,11 @@ function newApolloServer (app, schema, httpServer, tracing, playgroundConfig, gr
 
   let apolloServer = new ApolloServer({
     schema,
+    introspection: serverSecurity.introspection,
+    validationRules: [
+      depthLimit(serverSecurity.depthLimit),
```
darahayes: What happens if depthLimit is not a valid number?

server/config/index.js
StephenCoady: What about the case where the developer wants to disable introspection (for whatever reason) and run a dev environment?

server/apolloServer.js
StephenCoady: In the npm module source code it is wrapped in a try catch block where the try returns a validationContext. If there is an error returned by the catch instead then it gets ignored in this array by apollo and by graphql as well I am assuming.

TLDR: no depth validation but nothing breaks

server/config/index.js
darahayes: @StephenCoady that developer could run it in their dev environment with `NODE_ENV=production` 😄

darahayes:
@StephenCoady overall, this PR is good. The integration tests are failing for some strange reason. Would you be able to rebase your with the latest changes on master and we'll see if that gets the build passing?

server/config/index.js
StephenCoady: Ah ok, I didn't realise we weren't really doing anything else differently. Will push that now.

darahayes:
@StephenCoady I've pushed up an additional change that should make the output from the integration tests the smallest bit more readable. Might help us figure out what's going on.

StephenCoady:
@darahayes mind taking another look? fixed the tests.

PR #81, opened 2018-08-30
feat: support array of roles in keycloak hasRole

darahayes:
## Motivation

This PR adds the ability to specify an array of roles in the `hasRole` directive for keycloak. The following syntax is now supported:

```
@hasRole(role: ["editor", "admin", "reader"])
@hasRole(role: ["editor", "admin", "reader"], type: "client")
@hasRole(role: ["editor", "admin", "reader"], type: "realm")

@hasRole(role: "admin")
@hasRole(role: "admin", type: "client")
@hasRole(role: "admin", type: "realm")
```

## How

One of the main things I want to draw attention to is that I've introduced a new dependency called Joi. [Joi](https://npm.im/joi) is one of the most sophisticated validation libraries I've ever used.

You can see an example of how it's being used in the `validateArgs` function

```js
    const allowedTypes = ['client', 'realm']

    // joi is dope. Read the docs and discover the magic.
    // https://github.com/hapijs/joi/blob/master/API.md
    const argsSchema = Joi.object({
      role: Joi.array().required().items(Joi.string()).single(),
      type: Joi.string().valid(allowedTypes).default('client')
    })

    const result = argsSchema.validate(this.args)
```

If you're interested, here's quick summary of what the schema is doing:

* the input must be an object (`Joi.object`)
* The object must have the keys `role` and `type`. Any additional keys will cause a validation error. This means `@hasRole(role: "admin", type: "client", foo: "bar")` will cause an error because `foo` is not allowed.
* the role field is 
	* an array (`Joi.array()`), 
	* it is required (`.required()`) 
	* it must contain only strings (`.items(Joi.string())`)
	* if only a single string is given instead of an array, then it will be converted into an array with that string as the only element (`single()`). i.e. `{role: "admin"}` automatically becomes `{role: ["admin"]}`
* the type field
	* is a string
	* only the allowedTypes are valid (`.valid(allowedTypes)`)
	* if nothing is passed, the default value is `client`

Think about how much code you'd need to write to create that behaviour. This only scratches the surface in terms of what's possible with this library.

To learn more, check out the API docs: https://github.com/hapijs/joi/blob/master/API.md

I've been meaning to bring it in for some time but hadn't gotten around to it yet. There are plenty of other parts of the server where something like this would come in very handy. (validating resolvermappings, datasources, subscriptions etc at startup/reload time)

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

PS.: Add images and/or .gifs to illustrate what was changed if this pull request modify the appearance/output of something presented to the users.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4547,6 +4547,11 @@
         "minimalistic-crypto-utils": "1.0.1"
       }
     },
+    "hoek": {
+      "version": "5.0.4",
+      "resolved": "https://registry.npmjs.org/hoek/-/hoek-5.0.4.tgz",
+      "integrity": "sha512-Alr4ZQgoMlnere5FZJsIyfIjORBqZll5POhDsF4q64dPuJR6rNxXdDxtHSQq8OXRurhmx+PWYEE8bXRROY8h0w=="
+    },
     "hosted-git-info": {
       "version": "2.7.1",
       "resolved": "https://registry.npmjs.org/hosted-git-info/-/hosted-git-info-2.7.1.tgz",
@@ -5003,6 +5008,21 @@
       "resolved": "https://registry.npmjs.org/isarray/-/isarray-0.0.1.tgz",
       "integrity": "sha1-ihis/Kmo9Bd+Cav8YDiTmwXR7t8="
     },
+    "isemail": {
+      "version": "3.1.3",
+      "resolved": "https://registry.npmjs.org/isemail/-/isemail-3.1.3.tgz",
+      "integrity": "sha512-5xbsG5wYADIcB+mfLsd+nst1V/D+I7EU7LEZPo2GOIMu4JzfcRs5yQoypP4avA7QtUqgxYLKBYNv4IdzBmbhdw==",
+      "requires": {
+        "punycode": "2.1.1"
+      },
+      "dependencies": {
+        "punycode": {
+          "version": "2.1.1",
+          "resolved": "https://registry.npmjs.org/punycode/-/punycode-2.1.1.tgz",
+          "integrity": "sha512-XRsRjdf+j5ml+y/6GKHPZbrF/8p2Yga0JPtdqTIY2Xe5ohJPD9saDJJLPvp9+NSBprVvevdXZybnj2cv8OEd0A=="
+        }
+      }
+    },
     "isexe": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/isexe/-/isexe-2.0.0.tgz",
@@ -5059,6 +5079,16 @@
       "resolved": "https://registry.npmjs.org/jmespath/-/jmespath-0.15.0.tgz",
       "integrity": "sha1-o/Iiqarp+Wb10nx5ZRDigJF2Qhc=",
       "dev": true
+    },
+    "joi": {
+      "version": "13.6.0",
+      "resolved": "https://registry.npmjs.org/joi/-/joi-13.6.0.tgz",
+      "integrity": "sha512-E4QB0yRgEa6ZZKcSHJuBC+QeAwy+akCG0Bsa9edLqljyhlr+GuGDSmXYW1q7sj/FuAPy+ECUI3evVtK52tVfwg==",
+      "requires": {
+        "hoek": "5.0.4",
+        "isemail": "3.1.3",
+        "topo": "3.0.0"
+      }
     },
     "js-beautify": {
       "version": "1.7.5",
@@ -10131,6 +10161,14 @@
       "requires": {
         "is-number": "3.0.0",
         "repeat-string": "1.6.1"
+      }
+    },
+    "topo": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/topo/-/topo-3.0.0.tgz",
+      "integrity": "sha512-Tlu1fGlR90iCdIPURqPiufqAlCZYzLjHYVVbcFWDMcX7+tK8hdZWAfsMrD/pBul9jqHHwFjNdf1WaxA9vTRRhw==",
+      "requires": {
+        "hoek": "5.0.4"
       }
     },
     "toposort-class": {
--- a/package.json
+++ b/package.json
@@ -91,7 +91,8 @@
     "prom-client": "^11.1.1",
     "sequelize": "^4.38.0",
     "sequelize-cli": "^4.0.0",
-    "subscriptions-transport-ws": "^0.9.11"
+    "subscriptions-transport-ws": "^0.9.11",
+    "uuid": "^3.3.2"
   },
   "pre-commit": [
     "lint",
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "graphql-tools": "^3.0.2",
     "graphql-validation-complexity": "^0.2.3",
     "handlebars": "^4.0.11",
+    "joi": "^13.6.0",
     "json-parse-safe": "^1.0.5",
     "keycloak-connect": "^4.2.1",
     "lodash": "^4.17.10",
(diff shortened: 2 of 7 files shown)


wtrocki:
What I like is that you can also do 

`@hasRole(role: ["editor", "client2:admin"], type: "client") @hasRole(role: ["reader"], type: "realm")` 

and it will work.

wtrocki:
Update:
Together with @StephenCoady we found out that current implementation errors and this errors are not cached properly - as results users without role get's access to that resource. Additionally having types now seems to be confusing. From my personal point of view we could:

- Wrap entire code in try/catch to make sure that users will not get access to unauthenticated resource
- Drop type parameter as it's causing more problems now - users will follow the keycloak format.

This is up to discussion to pick something that will work best for our use case.

wtrocki:
To replicate issue https://gist.github.com/StephenCoady/a1f0bba4eb5827bb4d6a2277d98a8502#file-schema-L26

darahayes:
@wtrocki Using that schema, I get the following error when I try to do `createProfile`

```
{
  "data": null,
  "errors": [
    {
      "message": "Cannot read property 'clientId' of undefined",
      "locations": [
        {
          "line": 61,
          "column": 3
        }
      ],
      "path": [
        "createProfile"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "TypeError: Cannot read property 'clientId' of undefined",
            "    at hasRole (/Users/dara/mcp/data-sync-server/node_modules/keycloak-connect/middleware/auth-utils/token.js:80:13)",
            "    at Array.find (native)",
            "    at field.resolve (/Users/dara/mcp/data-sync-server/server/security/services/keycloak/schemaDirectives/hasRole.js:26:27)",
            "    at field.resolve (/Users/dara/mcp/data-sync-server/node_modules/graphql-extensions/dist/index.js:92:48)",
            "    at resolveFieldValueOrError (/Users/dara/mcp/data-sync-server/node_modules/graphql/execution/execute.js:531:18)",
            "    at resolveField (/Users/dara/mcp/data-sync-server/node_modules/graphql/execution/execute.js:495:16)",
            "    at /Users/dara/mcp/data-sync-server/node_modules/graphql/execution/execute.js:339:18",
            "    at /Users/dara/mcp/data-sync-server/node_modules/graphql/jsutils/promiseReduce.js:25:10",
            "    at Array.reduce (<anonymous>)",
            "    at promiseReduce (/Users/dara/mcp/data-sync-server/node_modules/graphql/jsutils/promiseReduce.js:22:17)",
            "    at executeFieldsSerially (/Users/dara/mcp/data-sync-server/node_modules/graphql/execution/execute.js:336:38)",
            "    at executeOperation (/Users/dara/mcp/data-sync-server/node_modules/graphql/execution/execute.js:289:55)",
            "    at executeImpl (/Users/dara/mcp/data-sync-server/node_modules/graphql/execution/execute.js:154:14)",
            "    at Object.execute (/Users/dara/mcp/data-sync-server/node_modules/graphql/execution/execute.js:131:35)",
            "    at Promise.resolve.then (/Users/dara/mcp/data-sync-server/node_modules/apollo-server-core/dist/runQuery.js:127:35)",
            "    at <anonymous>",
            "    at process._tickCallback (internal/process/next_tick.js:188:7)"
          ]
        }
      }
    }
  ]
}
```

However the original resolver never gets called and the user cannot get access to the actual query/mutation. Is this the same error or is there something else?

StephenCoady:
@darahayes there were two I was seeing. What do you get when you try `profile` ?

darahayes:
@StephenCoady the other error I see when I try `profile` is this:

```
{
  "data": null,
  "errors": [
    {
      "message": "Cannot read property 'content' of undefined",
      "locations": [
        {
          "line": 70,
          "column": 3
        }
      ],
      "path": [
        "profile"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "TypeError: Cannot read property 'content' of undefined",
            "    at hasRealmRole (/Users/dara/mcp/data-sync-server/node_modules/keycloak-connect/middleware/auth-utils/token.js:132:13)",
            "    at Array.find (native)",
            "    at field.resolve (/Users/dara/mcp/data-sync-server/server/security/services/keycloak/schemaDirectives/hasRole.js:24:27)",
            "    at field.resolve (/Users/dara/mcp/data-sync-server/node_modules/graphql-extensions/dist/index.js:92:48)",
            "    at resolveFieldValueOrError (/Users/dara/mcp/data-sync-server/node_modules/graphql/execution/execute.js:531:18)",
            "    at resolveField (/Users/dara/mcp/data-sync-server/node_modules/graphql/execution/execute.js:495:16)",
            "    at /Users/dara/mcp/data-sync-server/node_modules/graphql/execution/execute.js:364:18",
            "    at Array.reduce (<anonymous>)",
            "    at executeFields (/Users/dara/mcp/data-sync-server/node_modules/graphql/execution/execute.js:361:42)",
            "    at executeOperation (/Users/dara/mcp/data-sync-server/node_modules/graphql/execution/execute.js:289:122)",
            "    at executeImpl (/Users/dara/mcp/data-sync-server/node_modules/graphql/execution/execute.js:154:14)",
            "    at Object.execute (/Users/dara/mcp/data-sync-server/node_modules/graphql/execution/execute.js:131:35)",
            "    at Promise.resolve.then (/Users/dara/mcp/data-sync-server/node_modules/apollo-server-core/dist/runQuery.js:127:35)",
            "    at <anonymous>",
            "    at process._tickCallback (internal/process/next_tick.js:188:7)"
          ]
        }
      }
    }
  ]
}
```

Again, I never get access to the original resolver. I know exactly what the bug is but I'm trying to verify what @wtrocki said about, thrown errors resulting in the user getting access to the original resolver. I don't see this happening, I just get the error back. Is there some way that that this bug can lead to the user getting access to the actual query/mutation?

darahayes:
@wtrocki @StephenCoady The bug was because the original code was doing something like this:

```js
if (type === 'realm') {
  foundRole = roles.find(token.hasRealmRole)
} else {
  foundRole = roles.find(token.hasRole)
}
```

This was a very silly bug caused on my part. What happens here is that we're calling `Array.find()` which will loop through the array of roles and call the function provided, passing in the element (role) as the argument. In this case either `token.hasRealmRole` or `token.hasRole`. Just in case you're not familiar with this you can see the docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

Anyways, the issue is that the function you pass into `Array.find` will have its `this` object set to `undefined` and this is an issue because `hasRealmRole` and `hasRole` are probably doing things like checking for properties on the original token object by doing things like `this.clientId` or `this.roles` or whatever. Because `this` is undefined, an error is thrown.

I could modify the original code do do something like this:

```js
if (type === 'realm') {
  foundRole = roles.find(token.hasRealmRole.bind(token))
} else {
  foundRole = roles.find(token.hasRole.bind(token))
}
```

If you're not familiar with `bind()`, it takes the original function and gives you back the same one, where the `this` object is set to whatever argument you passed in. See the docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind

To me this is really obvious JS stuff, but I am conscious that `this` and things like `bind`, `call` and `apply` are tricky for lots of people so I've decided to make it simpler and write the following:

```js
if (type === 'realm') {
  foundRole = roles.find((role) => {
    return token.hasRealmRole(role)
  })
} else {
 foundRole = roles.find((role) => {
   return token.hasRole(role)
  })
}
```

darahayes:
@StephenCoady @wtrocki if you could try to verify again, that would be great.

PR #82, opened 2018-08-30
Update comment schema

danielpassos:
This PR:

* Remove comments from Meme (type and query)
* Add a new Query (comments by memeid)
* Add Owner relationship from Comment
* Fix Owner relationship on Meme (from Array to a single type)
* Fix responseMapping for all postgres querys
* Probably something else I can't remember right now :bowtie:

### DIFF
--- a/examples/memeolist.query.examples.graphql
+++ b/examples/memeolist.query.examples.graphql
@@ -54,7 +54,7 @@ query comments {
 }
 
 mutation postComment {
-  postComment(memeid: "Fha32ppqLNWEfX3b", comment: "Like it", owner: "Wojtek") {
+  postComment(memeid: "Fha32ppqLNWEfX3b", comment: "Like it", owner: "1") {
     id
     comment
     owner {
--- a/examples/memeolist.query.examples.graphql
+++ b/examples/memeolist.query.examples.graphql
@@ -9,11 +9,6 @@ query allMemes {
       displayname
       pictureurl
     }
-    comments {
-      id
-      comment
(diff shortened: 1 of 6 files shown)


wtrocki:
@danielpassos Changes looks great. I have no specific comment, but it's best to hold off as we do not have this replicated for IOS.

wtrocki:
Going to create ticket to do IOS update.

wtrocki:
let's rebase it later and get that in after demo together with android updates.

wtrocki:
Good stuff. Do we need to udpate proposal?
Let me known when you will be happy for me to check that again

danielpassos:
@wtrocki It's ready for review. I already have changed the Android and going to do the same on iOS.

wtrocki:
Good to be merged. I want to push that to community cluster for wider tests

danielpassos:
@wtrocki Just keep in mind if we update the cluster with it, the released apk will stop to work and there is no reason to have it anymore

darahayes:
@danielpassos no problems with this PR. Just one question. Why remove comments from the meme type?

danielpassos:
@darahayes From the memeolist app perspective

1. I don't know if the comment will be showed (you need to go to the comment screen), so why retrieve the comment with the meme and fire 1 query for each meme?
1. When a new comment is added it will not be displayed in any app without reloading all memes because comments were in the meme type/query

darahayes:
@danielpassos makes perfect sense. LGTM

PR #84, opened 2018-08-31
Bugfix: Server drops security when schema reloaded

wojta:
## Motivation
Server was dropping security when schema was reloaded.

## What
Some parameters are not passed in `server/server.js` in `newApolloServer()` and `newExpressApp` while schema is being reloaded.

## How
Pass the missing parameters.

## Verification Steps
1. Start server with Keycloak enabled
2. Call to `/graphql` endpoint invokes Keycloak login
3. Do schema update (using for example `data-sync-ui`)

Expected: Call to `/graphql` endpoint invokes Keycloak login (when not logged in)

## Checklist:

## Progress

- [x] Finished task

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -8,8 +8,8 @@ const SecurityService = require('./security/SecurityService')
 
 // middlewares
 const { log } = require('./lib/util/logger')
-const expressPino = require('express-pino-logger')({logger: log})
-const {getMetrics, responseLoggingMetric} = require('./metrics')
+const expressPino = require('express-pino-logger')({ logger: log })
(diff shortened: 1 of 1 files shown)


StephenCoady:
Good catch. I was talking to Dara about this yesterday and I don't think it makes sense to update the config in two separate places. Do you think it would be a good idea to maybe change all of these parameters to be one destructed object so we only need to change it in one place?

wojta:
Yeah, I'll improve the code a bit.

laurafitzgerald:
@darahayes @wojta is there a JIRA for the work required to do the simplification you mentioned?

PR #85, opened 2018-08-31
add log level

StephenCoady:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-7857


server/server.js
```diff
@@ -11,6 +11,10 @@ const { log } = require('./lib/util/logger')
 const expressPino = require('express-pino-logger')({logger: log})
 const {getMetrics, responseLoggingMetric} = require('./metrics')
 
+if (process.env.LOG_LEVEL) {
+  expressPino.logger.level = process.env.LOG_LEVEL
```
aliok: @StephenCoady This looks good, but I would just set it explicitly to some default if nothing is specified.
What do you think?

PR #88, opened 2018-09-05
Integration test with authentication

wojta:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-7840

## What
Write integration test for Sync server with Authentication (Keycloak)

## Why
There isn't any automated integration test yet.

## How
Authentication is managed through headless browser using [Puppeteer](https://github.com/GoogleChrome/puppeteer) after that authentication token is obtained and passed in headers for the Apollo client's requests 

## Checklist:

- [x] Code has been tested locally by PR requester
- [x] Changes have been successfully verified by another team member 

## Progress

- [x] Obtaining authentication request
- [x] Performing authenticated calls in some test
- [x] Check that when keycloak is enabled, it is not possible to make unauthenticated requests to the /graphql endpoint
- [x] Check that when the hasRole() directive is applied to a query/mutation that the user can only perform that query/mutation when they have the correct role. This must be checked for realm roles and client roles.
- [x] Check that when the hasRole directive is applied, a user without the appropriate role will be returned an error in the response. (Note that graphql does not use HTTP status codes, it always returns 200 and the error will be found in the response body)
- [x] Check that when the hasRole directive is applied to a type's field (e.g. date of birth on a user type) that users without the appropriate role cannot perform any queries/mutations while asking for that field. e.g. getUsers will work but if I ask for getUsers including date of birth, I get an error.
- [x] Check array multiple roles hasRole

## Additional notes
There is dependency to external running Keycloak instance on [https://keycloak.security.feedhenry.org/auth/](https://keycloak.security.feedhenry.org/auth/) with realm `Memeolist` and client roles `sync-server`
Several users were created in the realm with `test-` prefix, with client roles either already existing or created such as `admin`, `voter`


package.json
```diff
@@ -22,7 +22,9 @@
     "test": "ava '*.test.js' '**/*.test.js' '!integration_test/*'",
     "test:integration": "npm run db:init && ava --concurrency=1 integration_test/*.test.js",
     "test:cover": "nyc --reporter=text-lcov ava --concurrency=1 '*.test.js' '**/*.test.js' > coverage.lcov && codecov",
+    "test:inta": "npm run db:init && ava --concurrency=1 integration_test/auth.integration.test.js",
```
aliok: IMO, a longer name like `test:auth-integration` would be better.

aliok:
Verifying now.

package.json
```diff
@@ -22,7 +22,9 @@
...
     "start": "node ./index.js",
+    "debug": "node --nolazy --inspect-brk=9229 ./index.js",
```
aliok: @wojta Was this accidentally checked in?

jhellar:
Reviewed and verified, LGTM. Only thing I'm not sure about is if we need to get the access token via browser - is it also part of test? Browser automation may not be always stable. Can we just call Keycloak's endpoint to obtain the token? https://github.com/wojta/data-sync-server/pull/1

integration_test/util/auth.js
```diff
@@ -0,0 +1,35 @@
+const puppeteer = require('puppeteer')
+
+async function authenticateKeycloak (username, password) {
```
aliok: can we have some timeouts in the requests here? I happened to run the integration test in a poor network and it was waiting indefinitely (you know it goes to security.feedhenry.org).

integration_test/util/auth.js
```diff
@@ -0,0 +1,35 @@
...
+  const browser = await puppeteer.launch()
+  const page = await browser.newPage({args: ['--no-sandbox', '--disable-setuid-sandbox']})
+
+  async function loginProcess () {
```
aliok: When the password is wrong (imagine somebody changed the password on security.feedhenry.org), this method runs indefinitely. Maybe it is better to just fail if the login was not successful.

I was testing the password here to see what happens: https://github.com/aerogear/data-sync-server/pull/88/files#diff-6719cfa6ff7e6b161fc5f40bfe34efb7R54

aliok:
@wojta  I love this PR.  Also, it is the best PR description ever with a good plan. I left some minor comments but it is really good work.

One question: is it ok to use security.feedhenry.org in integration tests like this? We're exposing that we have these users with password `test123`:
* test-admin
* test-norole
* test-voter
* test-voter2
* test-realm-role

I don't have enough knowledge to understand if this would cause any problems. I also don't know if we're ok to use security.feedhenry.org like this.

cc @wei-lee

integration_test/auth.integration.test.js
```diff
@@ -0,0 +1,238 @@
...
+
+async function authenticate (test, username, password) {
+  test.log(`Authenticating as ${username}`)
+  const authHeaders = await auth.authenticateKeycloak(username, password)
+  context.helper.resetApolloClient(authHeaders)
+  test.log(`Authenticated as ${username}`)
+}
+
+test.before(async t => {
+  process.env.KEYCLOAK_CONFIG_FILE = require('path').resolve('./keycloak/keycloak.json')
```
wtrocki: We should not use the same realm for tests. Let's create separate realm as any changes and improvements can possibly break this test un future. With changes we have there if server goes down we are losing all the work. 

- We should have separate realm
- We should have separate client config

wojta: ok, good idea, I also don't know how this impacts security. I think we can use credentials passed by the environment into CircleCI and separate realm.

wojta:
@jhellar I was not aware of possibility to do it like that. I'll try your PR without browser.

package.json
wojta: yes

package.json
wojta: I won't remove it, but rename it to `test:integration:auth`. It's useful to run this part only.

integration_test/auth.integration.test.js
wtrocki: @wojta Not sure what you mean, but to answer this generally: 
we can use predefined users and map them into code in integration tests - there is no need for CircleCi to be aware of anything.

wojta:
@aliok Now you'll need to pass the password in the environment variable `KEYCLOAK_TEST_USER_PASSWORD` to execute Auth int. tests  . Test users are now in a different realm of our Keycloak.

aliok:
@wojta sounds good. I think this is better.

How about CircleCI though? I saw the integration tests passed there, but couldn't find where KEYCLOAK_TEST_USER_PASSWORD is defined.

wojta:
@wtrocki There is no user with realm admin role  in the new realm. Only `test-realm-role` has `test` role.

PR #91, opened 2018-09-05
fix: ensure directives are added

darahayes:
## Motivation

ensure the directives are actually loaded...


wojta:
Good catch, I was wondering why it won't accept my schema for the tests.

PR #92, opened 2018-09-05
Fix release pipeline

wtrocki:
## Motivation
 
Pass the right tag to release validator


wtrocki:
Alternative will be to change script to use CIRCLE_TAG here: https://github.com/aerogear/data-sync-server/blob/05b5e0a96a16787ab6083b56cf106f69cd453cf6/scripts/validateRelease.sh 

but I believe that this is a better approach

PR #94, opened 2018-09-07
added a format to the error message and logged it to console

StephenCoady:
## Motivation

The feedback item here: https://trello.com/c/VemGvOFs/118-20-as-a-developer-i-want-to-make-sure-that-all-user-provided-inputs-for-the-sync-server-such-as-schemas-request-mappings-and-res


server/apolloServer.js
```diff
@@ -13,7 +14,13 @@ function newApolloServer (app, schema, httpServer, tracing, playgroundConfig, gr
     schema,
     validationRules: [
       queryDepthLimit(serverSecurity.queryDepthLimit),
-      createComplexityLimitRule(serverSecurity.complexityLimit)
+      createComplexityLimitRule(serverSecurity.complexityLimit, {
+        formatErrorMessage: (cost) => {
+          const queryLog = 'query with cost ' + cost + ' exceeds complexity limit'
```
darahayes: two really minor nits

* `queryLog` is not the most descriptive name. Perhaps something like `complexityErrorMessage` would be more descriptive.
* es6 template strings are more desirable over string concatenation.

```
const complexityErrorMessage = `query with ${cost} exceeds complexity limit`
```

StephenCoady: I'll change the name. With the templating I actually had what you have suggested originally but our linter doesn't like that...

```
[eslint] Unexpected template string expression. (no-template-curly-in-string)
```

darahayes: You need to use backticks for template strings. e.g \`some string\` 

That linter error is basically saying 'hey it looks like you're trying to do a template string but you're using normal quotes'

StephenCoady: 🤦‍♂️

server/apolloServer.js
```diff
@@ -13,7 +14,13 @@ function newApolloServer (app, schema, httpServer, tracing, playgroundConfig, gr
     schema,
     validationRules: [
       queryDepthLimit(serverSecurity.queryDepthLimit),
-      createComplexityLimitRule(serverSecurity.complexityLimit)
+      createComplexityLimitRule(serverSecurity.complexityLimit, {
+        formatErrorMessage: (cost) => {
+          const errorMessage = `query with ${cost} exceeds complexity limit`
+          log.warn(errorMessage)
```
aliok: I haven't tested this but I have a question. Can the admin see what query caused this log? Do we print the query in another log message before this one?

StephenCoady: @aliok not in its current form. I will take a look at the module and see if its possible to log the context or something similar.

StephenCoady: @aliok I forgot about this sorry. I just ran it with createError instead of formatErrorMessage which gives me access to the node. However I cannot customise it much beyond that. What that means is a huge log as every field gets a before and after cost printed to the console. Do you think a 400 error and the custom error back to the client is enough in this case?

aliok: Ok, good point. We shouldn't be logging huge logs. That's a security risk itself vulnerable to flooding and DoS. Then it is fine as is.
Thanks @StephenCoady

aliok: err, ok, one other option is to use "DEBUG" log level for logging that huge log. What do you think?

StephenCoady: actually that would work. I'll make those changes now.

PR #95, opened 2018-09-10
fix: memeolist hasRole directive

psturc:
## Motivation
Fixes error 
```
Invalid hasRole directive on field <blahblah> on parent Mutation`
...
"type" is not allowed
```


darahayes:
Thanks for this fix @psturc 😃

PR #96, opened 2018-09-11
Experimental: knex vm resolvers

darahayes:
## Description

This PR is purely experimental. The only reason I created a PR is for visibility (in case anyone wants to try it out)

Right now it's pretty hacky here are the main points:

* removes the concept of writing SQL statements with handlebars
* Instead allows users to write some small amount of Javascript code and gives them access to the [`knex`](https://knexjs.org/#Schema) library for making db calls. This allows them to write resolvers like this:
screen shot 2018-09-11 at 12 33 26
* It's possible to do multiple queries by doing something like this:
screen shot 2018-09-11 at 12 42 55
* The code the user writes is run using a module called [vm2](https://github.com/patriksimek/vm2). vm2 is a wrapper around the native [vm](https://nodejs.org/api/vm.html) module included in the Node.js standard library. It adds some additional safeguards and security that the underlying module does not.
* Top level return statements are not allowed inside the user's code. Instead, the return value is always whatever was returned from the last statement in their code.
* when someone does something like `db.select().from('books')` the query is not actually being called. The return type of this statement is a knex builder object. The query is executed when you call `.then()`. We are calling `.then` outside of the vm right now.
* Basically, we can check whatever is returned from the script and if it has `.then()` we can call that and return the resulting value. This allows us to to some validation and error handling to ensure the user wrote a useful script.

### DIFF
--- a/server/lib/resolvers/builders/index.js
+++ b/server/lib/resolvers/builders/index.js
@@ -2,5 +2,6 @@
 module.exports = {
   // Matches the DataSource type ENUM in the models/dataSource.js
   InMemory: require('./nedb'),
-  Postgres: require('./postgres')
+  // Postgres: require('./postgres'),
+  Postgres: require('./knex')
 }
--- a/server/lib/resolvers/builders/index.js
+++ b/server/lib/resolvers/builders/index.js
@@ -2,6 +2,6 @@
 module.exports = {
   // Matches the DataSource type ENUM in the models/dataSource.js
   InMemory: require('./nedb'),
-  // Postgres: require('./postgres'),
-  Postgres: require('./knex')
+  Postgres: require('./postgres'),
+  Knex: require('./knex')
 }
--- a/server/lib/resolvers/builders/knex.js
+++ b/server/lib/resolvers/builders/knex.js
@@ -10,13 +10,6 @@ function buildKnexResolver (dataSource, compiledRequestMapping, compiledResponse
       const requestTimeStart = Date.now()
       info['dataSourceType'] = dataSource.type
 
-      // const queryString = compiledRequestMapping({
-      //   context: {
-      //     arguments: args,
-      //     parent: obj
-      //   }
-      // })
-
       const sandbox = {
         db: dataSourceClient,
         arguments: args,
@@ -27,8 +20,6 @@ function buildKnexResolver (dataSource, compiledRequestMapping, compiledResponse
         sandbox
       })
 
-      // const script = new VMScript(compiledRequestMapping)
-
       // result is whatever is implicitly returned from the script
       // in this case it will be the query builder object from knex
       const query = vm.run(compiledRequestMapping)
@@ -38,6 +29,7 @@ function buildKnexResolver (dataSource, compiledRequestMapping, compiledResponse
         log.info({ msg: 'result from knex', result })
 
         if (isEmpty(compiledResponseMapping)) {
+          auditLog(true, context.request, info, obj, args, null)
           return resolve(result)
         }
 
@@ -51,7 +43,6 @@ function buildKnexResolver (dataSource, compiledRequestMapping, compiledResponse
 
         auditLog(true, context.request, info, obj, args, null)
 
-        // const responseScript = new VMScript(compiledResponseMapping)
         return resolve(responsevm.run(compiledResponseMapping))
       }).catch((error) => {
         log.error({ msg: 'error from knex', error })
--- a/server/lib/resolvers/builders/knex.js
+++ b/server/lib/resolvers/builders/knex.js
@@ -4,25 +4,33 @@ const { updateResolverMetrics } = require('../../../metrics')
 const { isEmpty } = require('lodash')
 
 function buildKnexResolver (dataSource, compiledRequestMapping, compiledResponseMapping) {
+  const sandbox = {
+    db: dataSource.getClient()
+  }
+
+  const vm = new VM({
+    sandbox
+  })
+
+  // run the vm code which will return a function that executes the user code
+  // although the userCode is not actually being run inside the vm,
+  // it is bound to the context inside the vm so it cannot do things like require()
+  const userResolverFunction = vm.run(compiledRequestMapping)
+
   return function resolve (obj, args, context, info) {
     return new Promise((resolve, reject) => {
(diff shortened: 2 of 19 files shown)


wtrocki:
Added to my queue!

server/lib/resolvers/builders/index.js
```diff
@@ -2,5 +2,6 @@
 module.exports = {
   // Matches the DataSource type ENUM in the models/dataSource.js
   InMemory: require('./nedb'),
-  Postgres: require('./postgres')
+  // Postgres: require('./postgres'),
```
wtrocki: +1 to remove Postgres SQL resolvers as they are no providing standard security etc.

server/lib/resolvers/builders/knex.js
```diff
@@ -0,0 +1,60 @@
...
+        parent: obj
+      }
+
+      const vm = new VM({
+        sandbox
+      })
+
+      // result is whatever is implicitly returned from the script
+      // in this case it will be the query builder object from knex
+      const query = vm.run(compiledRequestMapping)
```
wtrocki: Let's add some logging here on debug level so it will provide more valuable tracing.

server/lib/resolvers/builders/knex.js
```diff
@@ -0,0 +1,60 @@
...
+
+      // this is a super rough implementation
+      query.then((result) => {
```
wtrocki: [Security] Does split between request/response still makes sense when we do not use sql?
With this split we have one operation outside sandbox (promise with query result), but if users return object different than promise with custom `then` method this method will be called outside sandbox.

I guess it depends if we want to really support kinda SaaS model here and it's not actual issue. Just wanted to point out.  

TL;DR - we may revisit request response format later.

server/lib/resolvers/builders/knex.js
```diff
@@ -0,0 +1,60 @@
...
+        log.info({ msg: 'result from knex', result })
+
+        if (isEmpty(compiledResponseMapping)) {
+          auditLog(true, context.request, info, obj, args, null)
+          return resolve(result)
+        }
+
+        const responseSandbox = {
+          result: result
```
wtrocki: So aproach will be in response users can only format existing result. 
Do you think we could support queries here as well?

server/lib/resolvers/resolverMapper.js
```diff
@@ -42,8 +42,22 @@ module.exports = function (dataSources, resolverMappings, pubsub) {
       throw new Error(`Resolver builder for ${dataSource.type} missing buildResolver function`)
     }
 
-    const compiledRequestMapping = compile(resolverMapping.requestMapping)
-    const compiledResponseMapping = compile(resolverMapping.responseMapping)
+    const compile = (dataSource.type === 'Postgres') ? compileScript : compileTemplate
```
wtrocki: If we remove SQL postgres we can improve this code later on.

server/lib/resolvers/resolverMapper.js
```diff
@@ -23,9 +23,9 @@ module.exports = function (dataSources, resolverMappings, pubsub) {
       throw new Error('Missing request mapping for mapping: ' + resolverMappingName)
     }
 
-    if (_.isEmpty(resolverMapping.responseMapping)) {
-      throw new Error('Missing response mapping for mapping: ' + resolverMappingName)
-    }
+    // if (_.isEmpty(resolverMapping.responseMapping)) {
```
wtrocki: The way I see it for 99% of the cases responseMapping will be empty (it will be always promise) so this makes me think that after merging this PR we may really review this implementation.

server/lib/resolvers/resolverMapper.js
wtrocki: However we can improve this code to make direct mapping between dataSource and method. For example
`
ResolverCompiler.compileFor(dataSource)
`

wtrocki:
From my point of view proposed change is definitely a way forward. Up to discussion:

- If request/response mappings are needed
- Is there value in old handlebars based template engine (for SQL and inmemory)
- Are we Sandboxing every request/response mappings etc.

darahayes:
Some Perf Notes:

## Handlebars Node v8.9.4

Running 10s test @ http://localhost:8000/graphql
10 connections

Stat         Avg     Stdev   Max
Latency (ms) 8.94    3.78    50.76
Req/Sec      1058.8  43.85   1117
Bytes/Sec    1.63 MB 72.1 kB 1.72 MB

11k requests in 10s, 16.3 MB read

## Handlebars Node v10.9.0

Stat         Avg     Stdev   Max
Latency (ms) 7.97    2.49    35.18
Req/Sec      1180.91 62.8    1266
Bytes/Sec    1.82 MB 96.3 kB 1.95 MB

12k requests in 10s, 18.2 MB read

## Knex Node v8.9.4

Stat         Avg     Stdev  Max
Latency (ms) 14.86   7.17   70.51
Req/Sec      650.3   20.74  682
Bytes/Sec    6.32 MB 240 kB 6.61 MB

7k requests in 10s, 63.2 MB read

## Knex Node v10.9.0

Stat         Avg     Stdev  Max
Latency (ms) 11.13   4.46   38.3
Req/Sec      860     24.96  910
Bytes/Sec    8.35 MB 310 kB 8.8 MB

9k requests in 10s, 83.2 MB read

wtrocki:
@darahayes `result[0]` in response mapper crashes when no result returned. 
Would it be possible to avoid response mapper

wtrocki:
When building queries I have noticed that saving query do not actually apply the change - when I referesh page previous version exist. However this is not happening on master.

wtrocki:
Crashes, errors are not propagated to application logs  - it's not possible to see any issues in application log

darahayes:
Hey @wtrocki yes the errors described are due to the fact that I changed the name of the knex datasource client and the resolver builder back to `Knex` (was `Postgres` initially). To get it to work you would have to create a new datasource of type `Knex`. The syntax error is due to the fact that the server is trying to send JavaScript to a postgres instance 😄 

In any case, this PR was moreso to demonstrate what is possible and it was never intended to be merged. Going to close this know but keep the branch in case we use the work later on.

wtrocki:
@darahayes Yes - Let's keep this branch. I'm basing on this work on other PR.

PR #97, opened 2018-09-13
feat: local Keycloak instance for integration tests

psturc:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-7906
* to have tests dependent only on locally running services - in this case, Keycloak
## What
Use local Keycloak instance for running integration tests
## Why
Why not
## How
* add keycloak image to `docker-compose.yml`
* add script for configuring local keycloak instance
* modify circleci.yml file so circleci will use local keycloak instance
## Verification Steps
* clone this branch
* run `docker-compose -p sync up` and wait for keycloak to spin up
* run `npm run test:integration:auth` (to test only auth) or `npm run test:integration` to run all integration tests
## Checklist:

## Progress

- [x] add exported Keycloak realm that suits the tests best (already contains realm roles & clients)
- [x] add script that takes care of Keycloak configuration (importing realm, creating users, assigning client & realm roles to them and deleting the realm once the test is finished)
- [x] remove `test.serialOrSkip` function & `KEYCLOAK_TEST_USER_PASSWORD` env var since they're not anymore necessary - for local keycloak instance, we can use any password and have it visible - currently it's set to `admin`
- [ ] modify [server/config/index.js](https://github.com/aerogear/data-sync-server/compare/AEROGEAR-7906?expand=1#diff-54ae0a40059ea9cc0549bf1c52f1a14a) to use [keycloak.json meant for testing](https://github.com/aerogear/data-sync-server/blob/master/integration_test/config/keycloak.json) when user is running running integration test
 

## Additional notes
You might ask: Why didn't he assigned the roles to a user when creating them?
[Click to show the answer.](https://stackoverflow.com/questions/49818453/client-roles-havent-assigned-during-creating-new-user-in-keycloak) I spend more than hour figuring out how to do that/finding workaround. So I decided to give up and do the user configuration afterwards.

### DIFF
--- a/server/config/index.js
+++ b/server/config/index.js
@@ -1,4 +1,3 @@
-
 const fs = require('fs')
 const { hostname } = require('os')
 
@@ -102,13 +101,10 @@ if (process.env.SCHEMA_LISTENER_CONFIG) {
 
 if (process.env.KEYCLOAK_CONFIG_FILE) {
   try {
-    const keycloakHost = process.env.KEYCLOAK_HOST || 'localhost'
-    const keycloakPort = process.env.KEYCLOAK_PORT || '8080'
     const keycloakConfig = fs.readFileSync(process.env.KEYCLOAK_CONFIG_FILE, 'utf-8')
 
     config.securityServiceConfig.type = 'keycloak'
     config.securityServiceConfig.config = JSON.parse(keycloakConfig)
-    config.securityServiceConfig.config['auth-server-url'] = `http://${keycloakHost}:${keycloakPort}/auth`
   } catch (ex) {
     log.error(`Unable to read keycloakConfig in ${process.env.KEYCLOAK_CONFIG_FILE} . Skipping it.`)
     log.error(ex)
--- a/server/config/index.js
+++ b/server/config/index.js
@@ -102,10 +102,13 @@ if (process.env.SCHEMA_LISTENER_CONFIG) {
 
 if (process.env.KEYCLOAK_CONFIG_FILE) {
   try {
+    const keycloakHost = process.env.KEYCLOAK_HOST || 'localhost'
+    const keycloakPort = process.env.KEYCLOAK_PORT || '8080'
     const keycloakConfig = fs.readFileSync(process.env.KEYCLOAK_CONFIG_FILE, 'utf-8')
 
     config.securityServiceConfig.type = 'keycloak'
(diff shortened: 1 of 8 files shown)


server/config/index.js
```diff
@@ -102,10 +102,13 @@ if (process.env.SCHEMA_LISTENER_CONFIG) {
 
 if (process.env.KEYCLOAK_CONFIG_FILE) {
   try {
+    const keycloakHost = process.env.KEYCLOAK_HOST || 'localhost'
+    const keycloakPort = process.env.KEYCLOAK_PORT || '8080'
     const keycloakConfig = fs.readFileSync(process.env.KEYCLOAK_CONFIG_FILE, 'utf-8')
 
     config.securityServiceConfig.type = 'keycloak'
     config.securityServiceConfig.config = JSON.parse(keycloakConfig)
+    config.securityServiceConfig.config['auth-server-url'] = `http://${keycloakHost}:${keycloakPort}/auth`
```
darahayes: Just a minor thing. This is a bit of a hack and would break existing functionality, requiring changes in the data-sync-apb. We should write/store a keycloak config file with the correct values inside the `integration_tests` folder and provide that to the server using `process.env.KEYCLOAK_CONFIG_FILE`

darahayes:
@psturc I left a small comment, but overall, this PR is amazing. Thank you so much for this 😃

darahayes:
@psturc I have one question, as part of the prepareKeycloak, we are creating some users. by making calls to the keycloak API. Is it possible to create some users + role, export the realm, reimport the realm in a different keycloak and have those users + roles still there? 

Or is creating the users via the API the only way?

server/config/index.js
psturc: Cool. I have it in my todo list and was just wondering how to achieve this. Rewriting the keycloak config file is good idea.

psturc:
@darahayes thanks for asking, I probably should have written it in the description. Basically when you do export (via Keycloak UI), you do partial export, which means almost everything except users and groups is exported.
[There is a way how to export users](https://www.keycloak.org/docs/latest/server_admin/index.html#_export_import) - it should land to a different file. I didn't find a way how to do that with dockerized version of Keycloak - approach described in the docs didn't work for me. To be honest, I'm not sure if it's worth investing the time into that :D

psturc:
@darahayes  all fixed now

darahayes:
Great work @psturc 😃

PR #98, opened 2018-09-18
Remove build schema

StephenCoady:
# Motivaton

This change will allow us to address some technical debt and architecture limitations.
Most important ones:
- Add ability to try out queries and schemas without publishing them to the server
- Add ability to Query schema when the server is protected
- Separate UI from Server (server will be production environment and UI will be development environment)
- Add support for additional use cases (user provided sync server can use our execution later)
- Improve and fix schema reload for sync
- Unify models that are being used in UI and Server (so they will be extended in one single place)
- Support schema seeding (memeo and other schemas - this is currently happening on Server - we need that ability on UI)

TL;DR - This is a crucial change that fixes some problems with the Sync architecture and it will give us tons of benefits with very little investment.

# Side effects of this change

- UI and Server will reuse the same module
- Development time will require using github version of new module (or releases to npm)
When we will get more and more modules we can consider moving towards mono repositories etc. but currently this impact is marginal

# Verification Steps

To verify that these changes work, it can be done as part of the overall verification of the feature. Using this repository and the sync UI.

1. Use docker-compose to create the db locally. 
1. Run the server with `npm run dev:keymemeo`
1. Run the UI with `npm run server`
1. Everything should work as expected with the server and the UI. i.e. authentication should be required on the server and not on the UI.


wtrocki:
# Motivation

This change will allow us to address some technical debt and architecture limitations.
Most important ones:
- Add ability to try out queries and schemas without publishing them to serevr
- Add ability to Query schema when server is protected
- Closely separate UI from Server (server will be production environment and UI will be development environment)
- Add support for additional use case (user provided sync server can use our execution later)
- Improve and fix schema reload for sync
- Unify models that are being used in UI and Server (so they will be extended in one single place)
- Support schema seeding (memeo and other schemas - this is currently happening on Server - we need that ability on UI)

TL;DR - This is the crucial change that fixes some problems with Sync architecture and it will give us tons of benefits with very little investments.

# Side effects of this change

- UI and Server will reuse the same module
- Development time will require using github version of new module (or releases to npm)
When we will get more and more modules we can consider moving towards mono repositories etc. but currently this impact is marginal

package.json
```diff
@@ -70,6 +70,7 @@
     "axios": "^0.18.0",
     "body-parser": "^1.18.3",
     "cors": "^2.8.4",
+    "data-sync-gql-core": "git+https://git@github.com/aerogear/data-sync-gql-core.git#master",
```
wtrocki: This works as well `aerogear/data-sync-gql-core#master`

server/server.js
```diff
@@ -18,8 +18,9 @@ if (process.env.LOG_LEVEL) {
 }
 
 // schema
-const buildSchema = require('./buildSchema')
+const buildSchema = require('data-sync-gql-core')
 const schemaListenerCreator = require('./lib/schemaListeners/schemaListenerCreator')
+const { makeExecutableSchema } = require('graphql-tools')
```
wtrocki: It feels strange that API do not include this. I guess this is happening because of version conflicts in the package and it's clearly technical debt that will be resolved later?

wtrocki:
@StephenCoady Could we get PR for executor as well.
The way we could do it is to:
- Push entire server master
- Remove non essential files directly on master
- Create PR with modifications and API.

It's best to hold of with this after @darahayes PR for knex will land into this module, however we can also apply this change later.

server/server.js
StephenCoady: Yes exactly. It works without me needing to include this, but there are multiple errors mentioning unexpected results. I will try to find a solution, but yes - technical debt.

StephenCoady:
@wtrocki mind taking another look? I have updated the core module to include the history now and this and the ui pr is updated. thanks.

server/server.js
```diff
@@ -74,7 +75,7 @@ class DataSyncServer {
     this.serverConfig = serverConfig
 
     // generate the GraphQL Schema
-    const { schema, dataSources } = await buildSchema(this.models, this.pubsub, this.serverConfig.schemaDirectives)
+    const { schema, dataSources } = await buildSchema(this.models, this.pubsub, this.serverConfig.schemaDirectives, makeExecutableSchema)
```
wtrocki: Actually I found out if we do not pass directives to graphql schema build it will fail validation. 
Currently we have @hasRole directive. The way I will suggest to aproach this is that core will have definitions for directives and server itself will provide implementations

sequelize/seeders/memeolist-example-shared.js
```diff
@@ -46,11 +46,11 @@ const memeoListSchema = {
...
     comments(memeid: ID!): [Comment]!
   }
-  
+    
   type Mutation {
-    createProfile(email: String!, displayname: String!, pictureurl: String!):Profile! @hasRole(role: "realm:admin")
+    createProfile(email: String!, displayname: String!, pictureurl: String!):Profile!
     createMeme(owner: ID!, photourl: String!):Meme!
-    likeMeme(id: ID!): Boolean @hasRole(role: "realm:voter")
+    likeMeme(id: ID!): Boolean @hasRole(role: "something")
```
wtrocki: Why this change?

server/server.js
wtrocki: This is resolved

StephenCoady:
@wtrocki this is ready for a final verification now I think. All tests are passing locally so like you said this is probably a race condition.

wtrocki:
Checking that and going to fix tests as well. @darahayes this changes are quite large so it's best to have more than one reviewer. Can we take look and verify this together?

server/server.js
```diff
@@ -18,20 +18,27 @@ if (process.env.LOG_LEVEL) {
...
     this.pubsub = pubsub
 
     this.schema = null
     this.dataSources = null
   }
 
+  async initializeCore (config) {
+    this.core = new Core(config.postgresConfig, makeExecutableSchema)
+    this.models = await this.core.getModels()
+    this.models.sync({ logging: false })
```
darahayes: @StephenCoady Just wondering, what's the rationale behind having this here? This is essentially like running a database migration every time the server starts up.

In master, we **purposefully avoid** calling this and it is up to the user to do initialize the database structure with `npm run db:init`. There are a couple of reasons for this:

* In local development, you might make manual changes to the db structure for whatever reason (new feature, debugging etc.) and this automatic sync would destroy those changes
* Without the automatic sync, it becomes very clear if there is any mismatch between the server and the db structure because the whole thing will crash.

Is there any specific reason this was needed here?

server/server.js
```diff
@@ -110,15 +117,15 @@ class DataSyncServer {
...
+    if (this.dataSources) await this.core.disconnectActiveDataSources(this.dataSources)
     if (this.server) await this.server.close()
   }
 
   async onSchemaChangedNotification () {
     log.info('Received schema change notification. Rebuilding it')
     let newSchema
     try {
-      newSchema = await buildSchema(this.models, this.pubsub, this.serverConfig.schemaDirectives)
+      newSchema = await this.core.buildSchema('default', this.pubsub, this.serverConfig.schemaDirectives)
```
darahayes: Just a minor thing: Can we declare this `'default'` value as a constant at the top of the file?

wtrocki: This reminds me that we can actually get schema name from pubsub :)

server/server.js
wtrocki: But this is only for metadata. Users should not actually manage our metadata storage.
Sync will just create required columns but our metadata format seems to be static.

wtrocki:
Verified changes. Looking into failed tests.

server/server.js
StephenCoady: This was an oversight on my part. Originally the server handled the sync of models at the service level before the server was started. I moved it but have moved it back now, will push shortly.

PR #99, opened 2018-09-21
fix: role name in package.json scripts

psturc:
## Why
`npm run db:init:memeo:postgres` ➡️  `postgres_memeo_1  | FATAL:  role "postgres" does not exist`


wtrocki:
Seems that current scripts are working fine. Closing as outdated.

PR #101, opened 2018-10-05
swapped keycloak port to 9090

StephenCoady:
Currently having Keycloak on 8080 was causing some problems with the UI (as this is the port it uses). Means we can't test everything at once.


StephenCoady:
@wtrocki could you take a look now?

wtrocki:
@StephenCoady Added small thing. Can you review this for me?

PR #103, opened 2018-10-08
Added script to initialize a local keycloak instance

ziccardi:
## Motivation
There were scripts to initialize the database to be used with the memeolist app, but no script was provided to configure a local keycloak.

## What
Added a new script to configure the local keycloak instance to be used with the memeolist app (heavily based on @psturc scripts)

## How
A new command is provided:

```bash
$ npm run keycloak:init:memeo

A local keycloak as been configured, you can access such instance at http://192.168.1.187.nip.io:8080/auth
username/password: admin/admin
The following users have been created: (username/password)

   * voter/voter
   * voter2/voter2

To use this keycloak instance with the syncserver, please use the following command:

KEYCLOAK_CONFIG_FILE=./keycloak/keycloak-local.json npm run dev
```

## Verification Steps

1. After running the `docker-compose` command as documented in the sync server docs, run the `npm run keycloak:init:memeo` command
2. run the sync server with `KEYCLOAK_CONFIG_FILE=./keycloak/keycloak-local.json npm run dev`
3. Inside the memeolist app, edit the `mobile-services.json` file to point to the provided keycloak URL
4. Run the memeolist and ensure that it works

## Checklist:

- [x] Code has been tested locally by PR requester

### DIFF
--- a/keycloak/index.js
+++ b/keycloak/index.js
@@ -1,0 +1,44 @@
+/* eslint no-console: 0 */
+const keycloakInit = require('./keycloakInit')
+const os = require('os')
+
+function getIpAddress () {
+  var ifaces = os.networkInterfaces()
+
+  for (var ifname of Object.keys(ifaces)) {
+    for (var iface of ifaces[ifname]) {
+      if (iface.family !== 'IPv4' || iface.internal !== false) {
+        // skip over internal (i.e. 127.0.0.1) and non-ipv4 addresses
+        continue
+      }
+
+      return iface.address
+    }
+  }
+}
+
+async function init () {
+  await keycloakInit.resetKeycloakConfiguration('http://127.0.0.1:8080/auth')
+  await keycloakInit.prepareKeycloak('http://127.0.0.1:8080/auth')
+
+  const keycloakUrl = `http://${getIpAddress()}.nip.io:8080/auth`
+  var json = `{ "realm": "Memeolist", "auth-server-url": "${keycloakUrl}", "ssl-required": "external", "resource": "sync-server", "public-client": true }`
+
+  var fs = require('fs')
+  fs.writeFile('./keycloak/keycloak-local.json', json, function (err) {
+    if (err) {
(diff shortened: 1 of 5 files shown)


keycloak/index.js
```diff
@@ -0,0 +1,44 @@
...
+      }
+
+      return iface.address
+    }
+  }
+}
+
+async function init () {
+  await keycloakInit.resetKeycloakConfiguration('http://127.0.0.1:8080/auth')
+  await keycloakInit.prepareKeycloak('http://127.0.0.1:8080/auth')
```
wtrocki: Why not use `getIpAddress` here?

wtrocki:
@ziccardi We had exported realm initially and had this for both developers and end users to be able to replicate this by importing this file:
https://github.com/aerogear/data-sync-server/blob/master/keycloak/realm-export.json3

This change introduces automatic loading, but to get that straight:
Are we going to also let users to use this way of seeding their container or it's just for developers?
From my point of view introduced changes are awesome however we may simply feed exported realm into this. Example here:
https://github.com/feedhenry-raincatcher/raincatcher-keycloak/blob/master/scripts/realm-init.sh

Before we jump into conclusion I would like to know if that is for developers/contributors only and we still need exported realm (and duplication) etc. What is your opinion about this?

darahayes:
@wtrocki I would argue this is mostly for development. It means that I can start up the entire data sync service with keycloak and have it preconfigured with the memeolist app, all without accessing external services.

wtrocki:
@darahayes Talked with @ziccardi and we seen problem of having two sources of the memeolist realm. One that is created in code and one that is actual export. Following code is extracted from integration tests that are not creating all clients (for all platforms) and missing some roles etc.
Realm based export could be documented in docs.

danielpassos:
@ziccardi Is this still valid?

darahayes:
@danielpassos this entire repo should probably be archived. Feel free to close the PR and do that.

PR #106, opened 2018-10-08
Update github files

camilamacedo86:


darahayes:
Hi @camilamacedo86 Thanks for the PR, The changes are greatly appreciated. I know this case is a bit of an exception, but for the sake of good process, please do not merge PRs unless they have at least one explicit approval in future :)

PR #107, opened 2018-10-08
Update github files - Remove license from GitHub dir

camilamacedo86:


camilamacedo86:
@aliok do you have any objection regards the files added here (.github dir ) and could we move forward with this one? We checked that the License should be not in the .github dir and the best place for it is the root of the project. Also, the biggest part of projects already has this file and this PR is just to remove it.

PR #109, opened 2018-10-23
aerogear-8024: use utils from the core package

aliok:
https://issues.jboss.org/browse/AEROGEAR-8024

Data sync server now accepts a `data-sync-client-info` header which is Base64 decoded and it prints whatever in that to audit logs.

This is to be done to have the client/app/device information in audit logs so that we can create aggregations/searches for the client info.

Work done:

* Adapt for the changes in https://github.com/aerogear/data-sync-gql-core/pull/14
* Return 400 if the client info is malformed.

How to verify:

1. Do the npm link stuff on core lib to have these changes in data-sync-server
2. Start data sync server
3. Open localhost:8000/graphql to access playground
4. Try querying (e.g. `{allMemes{id}}`) with these headers (can be added using the "HTTP HEADERS" button in the bottom):
4.1. No header --> no `clientInfo` field in audit logs
4.2. `{"data-sync-client-info":""}` --> no `audit.clientInfo` field in audit logs
4.3. `{"data-sync-client-info":"foo"}` --> should receive 400
4.4. `{"data-sync-client-info":"eyJhIjoxfQ=="}` --> `audit.clientInfo` field in audit logs with content `{"a":1}`
4.5. `{"data-sync-client-info":"eyJhIjoxfQfoo=="}` --> should receive 400
5. run `npm test`

Also note that the actual log is now wrapped in an `audit` field:
```
    tag: "AUDIT"
    audit: {
      "msg": "",
      "requestId": 34,
      "operationType": "query",
      "parentTypeName": "Query",
      "path": "allMemes",
      "success": true,
      "arguments": {},
      "dataSourceType": "InMemory",
      "clientInfo": {
        "a": 1
      }
    }
```


aliok:
Tests currently fail because of the unmerged changes in the core lib

server/lib/util/logger.js
```diff
@@ -1,35 +1,49 @@
 const log = require('pino')()
```
StephenCoady: we could use logger from core here, its the same thing.

wtrocki: -100.. Core logger is probably temporary there. For libraries people tend to use debug package

server/apolloServer.js
```diff
@@ -30,6 +33,15 @@ function newApolloServer (app, schema, httpServer, tracing, playgroundConfig, gr
       if (AuthContextProvider) {
         context.auth = new AuthContextProvider(req)
       }
+      if (auditLogEnabled) {
```
wtrocki: I personally preffer this to happen in the log module (so all logic will be there)
Then this code could look like this:
```
 context.clientInfo = req.clientInfo
context.auditLog = auditLog // Required from log (it can be NOP etc.)
```

Then this if is not executed for every request and we enclose all logging logic in this component. We do not have component based server but we could write stuff this way so if number of logic will grow we can extract that to separate elements easily.

server/expressApp.js
```diff
@@ -25,6 +29,24 @@ function newExpressApp (options, middlewares, securityService) {
...
   }
 
+  if (auditLogEnabled) {
+    app.use(function (req, res, next) {
+      try {
+        req.clientInfo = getClientInfoFromHeaders(req)
+      } catch (e) {
+        log.error('Error getting client info.')
+        log.error(e)
+        const message = 'Error decoding/parsing malformed client info. Message: ' + e.message
```
wtrocki: [optional] e.message could be exposed only on the development version. 
Generally best to not expose messages from errors. In this case we have control over this so it's not a big deal.

server/apolloServer.js
aliok: +1 will do

server/expressApp.js
aliok: {quote}
Any reason why we need to handle this try/catch here (we could use global error handler for express and attach error code there)
{quote}

That ends up in a 500 error, not 400.

wtrocki: @aliok yeah - so the common pattern is to supply status code to error object and global error handle reading status code and using it to send the message.
```javascript
const error = new Error("error msg")
error.statusCode = 400
error.payload= {...}
```

wtrocki:
@aliok Do you have sample curl query handy for testing?

wtrocki:
Working on verification

server/expressApp.js
aliok: @wtrocki 
But, in that case, don't we move the HTTP status related business logic into `getClientInfoFromHeaders` which is in `logger`?

Unless I misunderstood, I don't like doing this in `logger` module:
```
const error = new Error("error msg")
error.statusCode = 400
error.payload= {...}
```

wtrocki: @aliok We are moving towards talking about opinionated ways of handling errors. I do not have strong opinion so let's leave this untouched. 
Just pointed out that this is usually handled this way as it will allow us to enforce single format of the error payload etc.
Resolving problem above should be enough to get that merged IMHO.

wtrocki:
Verified. Let's merge core first in order for build to pass.

aliok:
@wtrocki give me your approval on https://github.com/aerogear/data-sync-gql-core/pull/14 then :)

server/expressApp.js
aliok: @wtrocki could you go a bit more in detail please? perhaps I didn't understand what you meant initially.

wtrocki: I think that your example clearly shows what I intended - throw error with all information that generic error handler can use to build.

aliok:
I don't get why the integration test started to fail... Seems unrelated.

aliok:
The integration test fails because Graphql was not accepting the case that a resolver with no representation in the schema existed.
This is either changed with my code (probably not), or changed with some dependency upgrade (thanks package-lock.json!)

Since NPM is basically ignoring the package-lock.json file, we either set the versions fixed in package.json or we upgrade and simply remove this test case.

@wtrocki @darahayes  wdyt?

And, npm crap: https://github.com/npm/npm/issues/18103#issuecomment-370401935

wtrocki:
Package was released  @aerogear/data-sync-gql-core (0.0.1) we can use that now and see if this is going to help/change anything

aliok:
@wtrocki it didn't change anything unfortunately. there's a behavior change with the upgraded version of Graphql, Apollo or something.
I think we're gonna have to upgrade eventually and this test case (not allowing a schema+resolvers when there is a missing entry in the schema for the resolver) will be obsolete. I will adapt the test.

PR #111, opened 2018-11-08
AEROGEAR-8094 - Sync server to extract user info from Keycloak token …

aliok:
JIRA: https://issues.jboss.org/browse/AEROGEAR-8094

Things done:
* If user is authenticated, print "authenticated:true" in the audit logs as well as the token details
* If user is not authenticated (also if auth is not enabled on sync server), print "authenticated:false"
* Created some abstraction for called token internals like `token.hasRole()`, `token.content`. So, instead of getting the token and calling 'token.hasRole()', leave it to 'context.auth.hasRole()'. This could be improved further by creating an interface like `AuthContextProvider` which could be implemented for different auth providers like `KeycloakAuthContextProvider`, etc.

Sample output with no auth enabled:
```
{
  tag: "AUDIT"
  audit: {
    "msg": "",
    "requestId": 1,
    "operationType": "query",
    "parentTypeName": "Query",
    "path": "allMemes",
    "success": true,
    "arguments": {},
    "dataSourceType": "InMemory",
    "authenticated": false
  }
}
```

Sample output with auth:
```
{
  tag: "AUDIT"
  audit: {
    "msg": "",
    "requestId": 19,
    "operationType": "query",
    "parentTypeName": "Query",
    "path": "allMemes",
    "success": true,
    "arguments": {},
    "dataSourceType": "InMemory",
    "authenticated": true,
    "userInfo": {
      "jti": "33981caa-f8a3-4527-9cfa-8d8d32540b72",
      "exp": 1541665472,
      "nbf": 0,
      "iat": 1541665172,
      "iss": "http://localhost:8080/auth/realms/Memeolist",
      "aud": "sync-server",
      "sub": "465879a1-30e8-446a-afd6-d8e99394b0ce",
      "typ": "Bearer",
      "azp": "sync-server",
      "auth_time": 1541664711,
      "session_state": "0c99d581-4f07-4873-8e90-7710026e396c",
      "acr": "0",
      "allowed-origins": [
        "localhost:8000"
      ],
      "realm_access": {
        "roles": [
          "admin",
          "uma_authorization",
          "voter"
        ]
      },
      "resource_access": {
        "account": {
          "roles": [
            "manage-account",
            "manage-account-links",
            "view-profile"
          ]
        }
      },
      "name": "Ali Ok",
      "preferred_username": "test",
      "given_name": "Ali",
      "family_name": "Ok",
      "email": "aliok@example.com"
    }
  }
}
```

How to verify:
1. Start sync server with memeo but no Keycloak and execute a query. Check the audit logs
2. Start sync server with memeo and Keycloak and execute a query. Check the audit logs.
```
cd keycloak 
docker-compose
go to localhost:8000
--> import realm-export.json into Keycloak
--> create a user with role voter or admin
--> reset pass for the user
--> change url in keycloak.json to http://localhost:8080/auth
npm run db:init:memeo:inmem
npm run dev:keymemeo
--> go to localhost:8080/graphql
--> login
--> go to localhost:8080/token and copy the token
--> paste it in the headers section in localhost:8080/graphql
--> execute a query
```

### DIFF
--- a/server/security/services/keycloak/AuthContextProvider.js
+++ b/server/security/services/keycloak/AuthContextProvider.js
@@ -9,6 +9,18 @@ class KeycloakAuthContextProvider {
     }
     return null
   }
+
+  isAuthenticated () {
+    return this.getToken() !== null
+  }
+
+  getTokenContent () {
+    return this.isAuthenticated() ? this.getToken().content : null
+  }
+
+  hasRole (role) {
+    return this.isAuthenticated() && this.getToken().hasRole(role)
+  }
 }
 
 module.exports = KeycloakAuthContextProvider
--- a/server/security/services/keycloak/AuthContextProvider.js
+++ b/server/security/services/keycloak/AuthContextProvider.js
@@ -1,17 +1,16 @@
 class KeycloakAuthContextProvider {
   constructor (request) {
     this.request = request
+    this.accessToken = (request && request.kauth && request.kauth.grant) ? request.kauth.grant.access_token : undefined
+    this.authenticated = !!(this.accessToken)
   }
 
   getToken () {
-    if (this.request.kauth && this.request.kauth.grant && this.request.kauth.grant.access_token) {
-      return this.request.kauth.grant.access_token
-    }
-    return null
+    return this.accessToken
   }
 
   isAuthenticated () {
-    return this.getToken() !== null
+    return this.authenticated
   }
 
   getTokenContent () {
--- a/server/security/services/keycloak/schemaDirectives/hasRole.js
+++ b/server/security/services/keycloak/schemaDirectives/hasRole.js
@@ -22,16 +22,21 @@ class HasRoleDirective extends SchemaDirectiveVisitor {
         throw newInternalServerError(context)
       }
 
-      const token = context.auth.getToken()
+      if (!context.auth || !context.auth.isAuthenticated()) {
+        const AuthorizationErrorMessage = `Unable to find authentication. Authorization is required for field ${field.name} on parent ${info.parentType.name}. Must have one of the following roles: [${roles}]`
+        log.error({ error: AuthorizationErrorMessage })
+        throw new ForbiddenError(AuthorizationErrorMessage)
+      }
+
       let foundRole = null // this will be the role the user was successfully authorized on
 
       foundRole = roles.find((role) => {
-        return token.hasRole(role)
+        return context.auth.hasRole(role)
       })
 
       if (!foundRole) {
         const AuthorizationErrorMessage = `user is not authorized for field ${field.name} on parent ${info.parentType.name}. Must have one of the following roles: [${roles}]`
-        log.error({ error: AuthorizationErrorMessage, details: token.content })
+        log.error({ error: AuthorizationErrorMessage, details: context.auth.getTokenContent() })
         throw new ForbiddenError(AuthorizationErrorMessage)
       }
 
--- a/server/lib/util/logger.js
+++ b/server/lib/util/logger.js
@@ -42,7 +42,9 @@ function auditLog (success, context, info, parent, args, msg) {
         parent: parent,
         arguments: args,
         dataSourceType: info.dataSourceType || '',
-        clientInfo: context ? context.clientInfo : undefined
+        clientInfo: context ? context.clientInfo : undefined,
+        authenticated: !!(context && context.auth && context.auth.isAuthenticated()),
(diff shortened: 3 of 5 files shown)


aliok:
@wei-lee 
```
"userInfo": {
      "jti": "33981caa-f8a3-4527-9cfa-8d8d32540b72",
      "exp": 1541665472,
      "nbf": 0,
      "iat": 1541665172,
      "iss": "http://localhost:8080/auth/realms/Memeolist",
      "aud": "sync-server",
      "sub": "465879a1-30e8-446a-afd6-d8e99394b0ce",
      "typ": "Bearer",
      "azp": "sync-server",
      "auth_time": 1541664711,
      "session_state": "0c99d581-4f07-4873-8e90-7710026e396c",
      "acr": "0",
      "allowed-origins": [
        "localhost:8000"
      ],
      "realm_access": {
        "roles": [
          "admin",
          "uma_authorization",
          "voter"
        ]
      },
      "resource_access": {
        "account": {
          "roles": [
            "manage-account",
            "manage-account-links",
            "view-profile"
          ]
        }
      },
      "name": "Ali Ok",
      "preferred_username": "test",
      "given_name": "Ali",
      "family_name": "Ok",
      "email": "aliok@aliok.com.tr"
    }
```

Can we print the token content above entirely? Or is there something in it that could create a security problem?

server/security/services/keycloak/AuthContextProvider.js
```diff
@@ -9,6 +9,18 @@ class KeycloakAuthContextProvider {
     }
     return null
   }
+
+  isAuthenticated () {
+    return this.getToken() !== null
```
darahayes: Minor thing. Contexts are created on every request. I've just realised how this code could be made simpler and more readable.

This class could be refactored such that `isAuthenticated` is a static value that gets initialised inside the constructor. e.g.

```js
this.isAuthenticated = (this.request.kauth && this.request.kauth.grant && this.request.kauth.grant.access_token)
```

Then we can have 


```js
isAuthenticated() {
	return this.isAuthenticated
}
```

Then `getToken` can get refactored such that:

```js
getToken() {
  if (this.isAuthenticated()) {
    return this.request.kauth.grant.access_token
  }
  return null
}
```

I think overall this is a bit simpler and it should be compatible with the new tests introduced in this PR. You don't have to do it but It might be worth a try :)

server/security/services/keycloak/schemaDirectives/hasRole.js
```diff
@@ -22,16 +22,21 @@ class HasRoleDirective extends SchemaDirectiveVisitor {
         throw newInternalServerError(context)
       }
 
-      const token = context.auth.getToken()
+      if (!context.auth || !context.auth.isAuthenticated()) {
+        const AuthorizationErrorMessage = `Unable to find authentication. Authorization is required for field ${field.name} on parent ${info.parentType.name}. Must have one of the following roles: [${roles}]`
```
darahayes: Right now, I don't see how this condition could occur, because if Keycloak is enabled, the GraphQL endpoint in the server is protected using a keycloak express middleware. 

This means that when keycloak is enabled, the server cannot be accessed unless the user is authenticated, so this code path **shouldn't be** reachable. 

That being said, it may enable a use case later on where it's possible to do certain things while authenticated/not authenticated. So +1 on this.

darahayes (approved):
Overall this is a great PR. Really like the additional tests 🎉

server/security/services/keycloak/AuthContextProvider.js
darahayes: @aliok feel free to take this or leave it. If these changes work without breaking the tests, it's an added bonus, but if the tests start failing, then I would scrap the idea for now 😃

aliok: +1 it makes sense. will check this out

server/security/services/keycloak/schemaDirectives/hasRole.js
aliok: > That being said, it may enable a use case later on where it's possible to do certain things while authenticated/not authenticated. So +1 on this.

yes, that was my thinking too.

server/security/services/keycloak/AuthContextProvider.js
aliok: done in https://github.com/aerogear/data-sync-server/pull/111/commits/0b0bce0e77aa86427f5e712e82468d55e53351ac

aliok:
@darahayes gonna merge this PR today

darahayes:
@aliok feel free to merge.

